PR #223, opened 2023-09-20
Update to typescript v5

SebastienGllmt:
Typescript v5 finally adds the `export type * from` notation. This will be extremely useful to publish the `paima-sdk` to NPM, so it was worth spending the time to upgrade

### DIFF
--- a/engine/paima-build-utils/src/middleware-esbuildconfig.template.cts
+++ b/engine/paima-build-utils/src/middleware-esbuildconfig.template.cts
@@ -1,6 +1,5 @@
 import { polyfillNode } from 'esbuild-plugin-polyfill-node';
 import type esbuild from 'esbuild';
-import { dtsPlugin } from 'esbuild-plugin-d.ts';
 import fs from 'fs';
 
 const define: Record<string, string> = { global: 'window' as const };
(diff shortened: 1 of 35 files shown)


engine/paima-build-utils/src/middleware-esbuildconfig.template.cts
```diff
@@ -39,11 +38,12 @@ const fsaReplace: esbuild.Plugin = {
...
 export const config: esbuild.BuildOptions = {
   // JS output from previous compilation step used here instead of index.ts to have more control over the TS build process
+  // however, that means type definitions will be lost until this is implemented: https://github.com/evanw/esbuild/issues/95#issuecomment-1559710310
   entryPoints: ['build/index.js'],
   bundle: true,
   format: 'esm',
   define,
   outfile: 'packaged/middleware.js',
-  plugins: [polyfillNode({}), dtsPlugin(), fsaReplace],
+  plugins: [polyfillNode({}), fsaReplace],
```
SebastienGllmt, line 47: This package doesn't work with typescript v5, and upon closer inspection this package actually did nothing anyway because the build script was changed from operation on `ts` files to operating on `js` files. I submitted a fix to chess so that we don't need this anymore anyway, and linked a ticket above that would be a cleaner fix

PR #224, opened 2023-09-22
Add evm chain if not in wallet.

acedward:
https://github.com/PaimaStudios/paima-engine/assets/3951895/d58992e5-059f-4fea-a872-8405dd2086d3

### DIFF
--- a/engine/paima-providers/src/evm.ts
+++ b/engine/paima-providers/src/evm.ts
@@ -161,17 +161,19 @@ export class EvmProvider implements IProvider<EvmApi> {
       const walletChain = await this.conn.api.request({ method: 'eth_chainId' });
       return parseInt(walletChain as string, 16) === parseInt(this.gameInfo.gameChainId, 16);
     } catch (e: any) {
-      throw new ProviderApiError(`[verifyWalletChain] error: ${e?.message}`);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      throw new ProviderApiError(`[verifyWalletChain] error: ${e?.message}`, e?.code);
     }
   };
   addChain = async (newChain: AddEthereumChainParameter): Promise<void> => {
     try {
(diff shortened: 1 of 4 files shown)


engine/paima-providers/src/evm.ts
```diff
@@ -161,17 +161,19 @@ export class EvmProvider implements IProvider<EvmApi> {
...
-      throw new ProviderApiError(`[verifyWalletChain] error: ${e?.message}`);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      throw new ProviderApiError(`[verifyWalletChain] error: ${e?.message}`, e?.code);
     }
   };
   addChain = async (newChain: AddEthereumChainParameter): Promise<void> => {
     try {
       await this.conn.api.request({
-        method: 'eth_chainId',
+        method: 'wallet_addEthereumChain',
```
SebastienGllmt, line 171: Oops, my bad. This was probably a copy-paste mistake

On a related note, this function if called by `async function switchChain()`, but I realize now that the function also never calls `switchChain` after it adds the network which maybe causes some issues in some wallets

engine/paima-providers/src/evm.ts
```diff
@@ -199,7 +202,8 @@ export class EvmProvider implements IProvider<EvmApi> {
       }
       return hash;
     } catch (e: any) {
-      throw new ProviderApiError(`[switchChain] error: ${e?.message}`);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      throw new ProviderApiError(`[switchChain] error: ${e?.message}`, e?.code);
```
SebastienGllmt, line 206: I noticed I made a copy-paste mistake here too. The error should be `sendTransaction` and not `switchChain`

SebastienGllmt (approved):

PR #228, opened 2023-10-04
Fix OS specific bash env

aljosa:
Here is a PR which fixes bash env for NixOS (also tested on macOS).

Not sure it's important enough to merge but maybe it will be useful to other people.


SebastienGllmt (approved):
LGTM

PR #246, opened 2023-10-24
add carp funnel with stake delegation pool tracking

ecioppettini:
## TODO

- [x] remove tmp_carp_client and use the version from npm once published
- [x] make the presync work
- [x] parametrize the known shelley time or something (potentially the carp service could provide this instead)

## Example settings

Currently I'm setting things up like this:

### extensions.yml

```
extensions:
   - name: "Cardano Stake Delegation"
     type: cardano-stake-delegation 
     pools: ["367d927a91eb9ecf15648779beeec8c710171566edcce18feb048265"]
     startSlot: 32032590  
     scheduledPrefix: cd

```

### .env (example)

```
CARP_URL=http://localhost:3000
CARDANO_NETWORK=preview
CARDANO_CONFIRMATION_DEPTH=10 
BATCHER_CARDANO_ENABLED_POOLS="095dd39da2d8534f9ddb93759a1931288e5dd79ae04fa5914e157bd6,093de39da2d8534f9ddb93759a1931288e5dd79a404fa5914e157b99"
BATCHER_CARP_URL=http://host.docker.internal:3000
```

Ideally the carp setting should be a single one, but I needed this to make it work with the batcher running in docker.

### Parsing

I had to setup a parsing rule like this:

```
cardanoDelegation  = cd|address|pool
```

(the lenghts are not tight bounds, just as an example for testing)

```
const cardanoDelegation: ParserRecord<CardanoStakeDelegation> = {
  address: PaimaParser.NCharsParser(0, 100),
  pool: PaimaParser.OptionalParser(null, PaimaParser.NCharsParser(0, 100)),
};
```


packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -0,0 +1,269 @@
...
+const confirmationDepth = '10';
+
+function knownTime(): number {
+  switch (ENV.CARDANO_NETWORK) {
+    case 'preview':
+      return 1666656000;
+    case 'preprod':
+      return 1666656000;
+    case 'mainnet':
+      return 1666656000;
```
gostkin, line 37: 1596059091

tmp-carp-client/README.md
SebastienGllmt, line 1: Just a reminder we'll have to remove all of these before we merge

packages/paima-sdk/paima-utils/src/config.ts
```diff
@@ -105,4 +105,13 @@ export class ENV {
   static get BATCHER_URI(): string {
     return process.env.BATCHER_URI || '';
   }
+
+  // TODO: put this in the right place
```
SebastienGllmt, line 109: TODO

packages/engine/paima-runtime/src/types.ts
```diff
@@ -119,30 +127,40 @@ export interface CdeErc6551RegistryDatum extends CdeDatumBase {
...
 
 const EvmAddress = Type.Transform(Type.RegExp('0x[0-9a-fA-F]{40}'))
   .Decode(value => value.toLowerCase())
   .Encode(value => value);
 
-const ChainDataExtensionConfigBase = Type.Object({
-  name: Type.String(),
-  startBlockHeight: Type.Number(),
-});
+const ChainDataExtensionConfigBase = Type.Union([
```
SebastienGllmt, line 158: Why is this now a union of a single object?

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -62,94 +63,115 @@ async function runPresync(
...
+      Network,
+      number
+    >;
   }
-  const filteredPresyncDataList = latestPresyncDataList.filter(
-    unit => unit.extensionDatums.length > 0
-  );
+
+  const filteredPresyncDataList = Object.values(latestPresyncDataList)
+    .flatMap(data => (data !== 'finished' ? data : []))
```
SebastienGllmt, line 164: Again, avoid magic strings

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -62,94 +63,115 @@ async function runPresync(
...
+    .filter(unit => unit.extensionDatums.length > 0);
   for (const presyncData of filteredPresyncDataList) {
     await gameStateMachine.presyncProcess(chainFunnel.getDbTx(), presyncData);
   }
-  return toBlock + 1;
+
+  return Object.fromEntries(
+    from
+      .filter(arg => latestPresyncDataList[arg.network] !== 'finished')
```
SebastienGllmt, line 172: Again, avoid magic strings

packages/engine/paima-sm/src/index.ts
```diff
@@ -73,13 +88,29 @@ const SM: GameStateMachineInitializer = {
...
-          latestCdeData.extensionDatums,
-          dbTx
-        );
-        if (cdeDataLength > 0) {
-          doLog(`Processed ${cdeDataLength} CDE events in block #${latestCdeData.blockNumber}`);
+        if (latestCdeData.network == Network.EVM) {
+          const cdeDataLength = await processCdeData(
+            latestCdeData.blockNumber,
+            latestCdeData.extensionDatums?.filter(
+              // TODO: find a more generic way of filtering
```
SebastienGllmt, line 95: TODO

packages/engine/paima-sm/src/index.ts
```diff
@@ -73,13 +88,29 @@ const SM: GameStateMachineInitializer = {
...
+              data => data.cdeDatumType == ChainDataExtensionDatumType.CardanoPool
+            ),
+            dbTx
+          );
+          if (cdeDataLength > 0) {
+            doLog(`Processed ${cdeDataLength} CDE events in block #${latestCdeData.blockNumber}`);
+          }
+        }
+        // TODO: duplicated code?
```
SebastienGllmt, line 104: TODO

packages/engine/paima-sm/src/index.ts
```diff
@@ -73,13 +88,29 @@ const SM: GameStateMachineInitializer = {
...
       },
       presyncProcess: async (dbTx: PoolClient, latestCdeData: PresyncChainData): Promise<void> => {
-        const cdeDataLength = await processCdeData(
-          latestCdeData.blockNumber,
-          latestCdeData.extensionDatums,
-          dbTx
-        );
-        if (cdeDataLength > 0) {
-          doLog(`Processed ${cdeDataLength} CDE events in block #${latestCdeData.blockNumber}`);
+        if (latestCdeData.network == Network.EVM) {
```
SebastienGllmt, line 91: Use === in Javascript

packages/engine/paima-sm/src/index.ts
```diff
@@ -73,13 +88,29 @@ const SM: GameStateMachineInitializer = {
...
+              data => data.cdeDatumType == ChainDataExtensionDatumType.CardanoPool
+            ),
+            dbTx
+          );
+          if (cdeDataLength > 0) {
+            doLog(`Processed ${cdeDataLength} CDE events in block #${latestCdeData.blockNumber}`);
+          }
+        }
+        // TODO: duplicated code?
+        else if (latestCdeData.network == Network.CARDANO) {
```
SebastienGllmt, line 105: Use ===

packages/engine/paima-sm/src/index.ts
```diff
@@ -73,13 +88,29 @@ const SM: GameStateMachineInitializer = {
...
-          dbTx
-        );
-        if (cdeDataLength > 0) {
-          doLog(`Processed ${cdeDataLength} CDE events in block #${latestCdeData.blockNumber}`);
+        if (latestCdeData.network == Network.EVM) {
+          const cdeDataLength = await processCdeData(
+            latestCdeData.blockNumber,
+            latestCdeData.extensionDatums?.filter(
+              // TODO: find a more generic way of filtering
+              data => data.cdeDatumType == ChainDataExtensionDatumType.CardanoPool
```
SebastienGllmt, line 96: Why would this be true in the `== EVM` case?

packages/engine/paima-sm/src/index.ts
```diff
@@ -188,27 +221,62 @@ async function processCdeData(
...
         await query.run(params, dbTx);
       }
-      await markCdeDatumProcessed.run(
-        {
-          block_height: blockHeight,
-          datum_count: datumCount,
-        },
-        dbTx
-      );
+      // await markCardanoCdeDatumProcessed.run({ slot: h, datum_count: datumCount }, dbTx);
```
SebastienGllmt, line 224: Do we need this commented out code?

packages/paima-sdk/paima-db/migrations/up.sql
```diff
@@ -28,6 +28,12 @@ CREATE TABLE cde_tracking (
   done BOOLEAN NOT NULL DEFAULT false
 );
 
+CREATE TABLE cde_tracking_cardano (
+  slot INTEGER PRIMARY KEY,
+  datum_count INTEGER NOT NULL,
```
SebastienGllmt, line 33: Why do we need to store the datum count? When would we only be part-way through tracking a slot?

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -0,0 +1,269 @@
...
+    default:
+      throw new Error('unknown cardano network');
+  }
+}
+
+function timestampToAbsoluteSlot(timestamp: number): number {
```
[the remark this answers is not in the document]
ecioppettini: do we care about byron though? With delegation it doesn't make sense afaik. Is there a case we we may need to consider pre-shelley blocks?

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -0,0 +1,269 @@
...
+      }),
+      DEFAULT_FUNNEL_TIMEOUT
+    );
+
+    if (stableBlock.block.slot > max) {
+      break;
+    }
+
+    await delay(1000);
+  }
```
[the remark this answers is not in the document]
ecioppettini: It can apply to syncing, but also a block may not have been produced yet. 

This is here in case the block at depth `confirmationDepth - 1` needs to be included in the range according to its timestamp/slot, but until an extra block is added at the tip then we don't want to consider it. This is because confirmation depends in depth, which is not necessarily related to time/slots.

Not sure how much it will matter in practice though, since it depends on the speed of both chains, but just in case.

Also, I guess the `1000` could be a constant.

packages/paima-sdk/paima-db/migrations/up.sql
ecioppettini: is this any different than the evm events? There can be more than one tx per block, so there can be more than one delegation for example, and I guess something could go wrong in between? I mostly just replicated the way it works with the existing cde's, since it didn't look any different to me

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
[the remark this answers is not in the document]
SebastienGllmt: Okay, we should have a comment to explain this

packages/paima-sdk/paima-db/migrations/up.sql
SebastienGllmt: Oh you're right - this was dead code that I didn't remove. I fixed this now: https://github.com/PaimaStudios/paima-engine/pull/257

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -57,94 +56,123 @@ async function runPresync(
...
   );
 
-  if (presyncBlockHeight > maximumPresyncBlockheight) {
-    // doLog(
-    //   `[paima-runtime] Skipping presync (presync block height: ${presyncBlockHeight}, maximum presync block height: ${startBlockHeight})`
-    // );
-    return;
+  if (!ENV.CARP_URL) {
+    delete presyncBlockHeight[Network.CARDANO];
   }
```
SebastienGllmt, line 67: Could this not cause a subtle issue where if you specify a Cardano CDE but the CARP_URL is missing that it just silently doesn't track anything? I feel if you're using task that requires Carp and you didn't specify the CARP_URL it should be an error

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -57,94 +56,123 @@ async function runPresync(
...
-        upper
+        Object.entries(presyncBlockHeight).map(([network, height]) => ({
+          network: Number(network),
+          from: height,
+          to: upper[network],
+        }))
       );
     });
+
+    presyncBlockHeight = await newHeight;
```
SebastienGllmt, line 100: Is there a reason the await is on 2nd line like this?

packages/engine/paima-rest/src/tsoa/routes.ts
```diff
@@ -1,16 +1,7 @@
 /* tslint:disable */
```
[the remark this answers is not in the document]
ecioppettini: deleted

packages/engine/paima-runtime/src/runtime-loops.ts
ecioppettini: yeah, makes sense, changed it for an error

packages/engine/paima-runtime/src/runtime-loops.ts
ecioppettini: there was a reason at some point before, but not anymore

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -0,0 +1,269 @@
...
+import { composeChainData, groupCdeData } from '../../utils';
+import { BaseFunnel } from '../BaseFunnel';
+import type { FunnelSharedData } from '../BaseFunnel';
+import type { PoolClient } from 'pg';
+import getCdePoolData from '../../cde/cardanoPool';
+import { Network } from '@paima/utils/src/constants';
+import { query, getErrorResponse } from '@dcspark/carp-client/client/src/index';
+import { Routes } from '@dcspark/carp-client/shared/routes';
+
+const confirmationDepth = '10';
```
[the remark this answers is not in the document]
ecioppettini: I made it an env... Since I think it should be configurable, although it's a bit problematic since it shouldn't really be changed without resyncing.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -0,0 +1,301 @@
...
+
+          return data;
+        default:
+          return Promise.resolve([]);
+      }
+    })
+  );
+
+  let grouped = groupCdeData(
+    Network.EVM,
```
gostkin, line 267: why not cardano?

gostkin:
how do I get pool id to put in the config in `BATCHER_CARDANO_ENABLED_POOLS`? what is the format? e.g. I have pool id `3867a09729a1f954762eea035a82e2d9d3a14f1fa791a022ef0da242`, how to put it there? `8200581ceeef9389ba6f3971425a59af99c56d1c9c227b3a2c2c8f52a10dd8ag` looks like a different format and i can't query it in the explorer

ecioppettini:
> how do I get pool id to put in the config in `BATCHER_CARDANO_ENABLED_POOLS`? what is the format? e.g. I have pool id `3867a09729a1f954762eea035a82e2d9d3a14f1fa791a022ef0da242`, how to put it there? `8200581ceeef9389ba6f3971425a59af99c56d1c9c227b3a2c2c8f52a10dd8ag` looks like a different format and i can't query it in the explorer

The `8200581ceeef9389ba6f3971425a59af99c56d1c9c227b3a2c2c8f52a10dd8ag` format was a mistake actually, it should be a string of 56 chars. I corrected it in the docs, but I forgot to do it here. It should be what shows as pool id in cardanoscan.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
ecioppettini: well... It doesn't actually matter, I think. 

The discriminant was introduced so that during the presync we can update the cardano cde tracking table instead of the other one, but during the normal sync just knowing the EVM block gives you the slot, so the field is not used.

But I guess we should put cardano here anyway to be consistent.

packages/engine/paima-runtime/src/cde-config/utils.ts
```diff
@@ -1,15 +1,37 @@
...
+    .map(cde => {
+      if (cde.cdeType != ChainDataExtensionType.CardanoPool) {
+        return cde.startBlockHeight;
+      } else {
+        return null;
+      }
+    })
+    .filter(sbh => !!sbh) as number[];
   const minStartBlockheight = Math.min(...startBlockheights);
   return isFinite(minStartBlockheight) ? minStartBlockheight : -1;
```
SebastienGllmt, line 18: All of this should probably be a single `.reduce` call instead (same with the function below)

packages/engine/paima-runtime/src/cde-config/utils.ts
```diff
@@ -1,15 +1,31 @@
...
 export function getEarliestStartBlockheight(config: ChainDataExtension[]): number {
-  const startBlockheights = config.map(cde => cde.startBlockHeight).filter(sbh => !!sbh);
+  const startBlockheights = config.reduce((arr, cde) => {
+    if (cde.cdeType !== ChainDataExtensionType.CardanoPool) {
+      arr.push(cde.startBlockHeight);
+    }
+    return arr;
+  }, [] as number[]);
   const minStartBlockheight = Math.min(...startBlockheights);
   return isFinite(minStartBlockheight) ? minStartBlockheight : -1;
```
SebastienGllmt, line 15: we can put more in the reduce. There isn't a point in building an array just to call `Math.min` on it afterwards (instead of just having the reduce find the minimum)

packages/engine/paima-funnel/src/cde/reading.ts
```diff
@@ -21,13 +21,15 @@ export async function getUngroupedCdeData(
...
+    extensions.map(extension =>
+      'startBlockHeight' in extension ? getSpecificCdeData(extension, fromBlock, toBlock) : []
+    )
   );
   return allData;
 }
 
 async function getSpecificCdeData(
-  extension: ChainDataExtension,
+  extension: ChainDataExtension & { startBlockHeight: number },
```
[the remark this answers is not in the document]
ecioppettini: I don't understand this question, it's used in the first line of the function actually

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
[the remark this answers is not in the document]
ecioppettini: By `AbsoluteSlot` I mean the cardano slot relative to the start of the chain, and not to the epoch. Which sometimes is also just called `slot` (like in carp).

`knownTime`  (now called `knownShelleyTime`) is the unix timestamp of the first block of shelley, which in the case of preview for example is also the genesis, but for mainnet is now.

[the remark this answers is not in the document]
ecioppettini: Also, answering my own question

>  Is there a case we we may need to consider pre-shelley blocks?

This is not used during the presync, it only matters during the sync, in order to map the EVM block to a unique slot in cardano, so that wouldn't be a problem anyway.

SebastienGllmt (approved):

PR #259, opened 2023-11-28
Projected NFT funnel

gostkin:
Projected NFT Funnel

### Setup

* Add `extensions.yml` to folder where the game binary is located:
```yaml
extensions:
  - name: "Cardano Projected NFT"
    type: cardano-projected-nft
    contract: "703ec99926211daafe291951b778d4da7736a8b0da2aeb92305b511a72"
    startSlot: 43201686
    scheduledPrefix: cpnft
```
* Add vars to `.env`:
```
CARP_URL=http://localhost:3000
CARDANO_NETWORK=preprod
CARDANO_CONFIRMATION_DEPTH=10
```


packages/engine/paima-funnel/src/cde/cardanoProjectedNFT.ts
```diff
@@ -0,0 +1,67 @@
...
+    query(url, Routes.projectedNftEventsRange, {
+      range: { minSlot: fromAbsoluteSlot, maxSlot: toAbsoluteSlot },
+    }),
+    DEFAULT_FUNNEL_TIMEOUT
+  );
+
+  return events
+    .map(e => eventToCdeDatum(e, extension, getBlockNumber(e.actionSlot)))
+    .filter(e => e != null)
+    .map(e => e!);
```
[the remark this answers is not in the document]
gostkin: why?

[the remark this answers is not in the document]
SebastienGllmt: @ecioppettini `e!` is special typescript syntax to ignore the nullability of a variable. Generally for cases like this I prefer to use reduce since it gives better type inference (so you don't have to use `e!`)

packages/engine/paima-sm/src/types.ts
```diff
@@ -71,6 +71,23 @@ interface CdeDatumCardanoPoolPayload {
...
+  ownerAddress: string;
+
+  actionTxId: string;
+  actionOutputIndex: number | undefined;
+
+  previousTxHash: string | undefined;
+  previousTxOutputIndex: number | undefined;
+
+  asset: string;
+  amount: number;
```
SebastienGllmt, line 84: This is wrong. Assets are of type `uint` in the CDDL spec for Cardano which can be up to 64 bits (larger than Javascript's max 53 bits)

JSON allows for infinite precision so JSON is fine, but any time you parse JSON to generate JSON in Javascript you lose precision if it's more than 53 bits.

packages/engine/paima-funnel/src/cde/cardanoProjectedNFT.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return {
+    cdeId: extension.cdeId,
+    cdeDatumType: ChainDataExtensionDatumType.CardanoProjectedNFT,
+    blockNumber,
+    payload: {
+      ownerAddress: event.ownerAddress != null ? event.ownerAddress : '',
+
+      actionTxId: event.actionTxId,
```
SebastienGllmt, line 51: Note that games only care about NFTs getting locked and queued to unlock. We don't care if the user ever actually ends up withdrawing their NFT from the lock later.

packages/engine/paima-sm/src/types.ts
```diff
@@ -262,6 +288,20 @@ export type ChainDataExtensionCardanoDelegation = ChainDataExtensionBase &
     cdeType: ChainDataExtensionType.CardanoPool;
   };
 
+export const ChainDataExtensionCardanoProjectedNFTConfig = Type.Object({
+  type: Type.Literal(CdeEntryTypeName.CardanoProjectedNFT),
+  contract: Type.String(),
```
SebastienGllmt, line 293: Probably this should be more description (ex: `scriptHash`)

packages/node-sdk/paima-utils-backend/src/cde-access.ts
```diff
@@ -167,3 +169,13 @@ export async function getCardanoAddressDelegation(
 ): Promise<string | null> {
   return await internalGetCardanoAddressDelegation(readonlyDBConn, address);
 }
+
+/**
+ * Fetch the pool this address is delegating to, if any.
```
SebastienGllmt, line 174: Comment needs to be updated

packages/engine/paima-sm/src/types.ts
```diff
@@ -262,6 +288,20 @@ export type ChainDataExtensionCardanoDelegation = ChainDataExtensionBase &
...
+  scheduledPrefix: Type.String(),
```
SebastienGllmt, line 294: Probably this should be option since most games will probably not need to update their state machine based on somebody locking or queuing to unlock

Probably what is more important is the `getCardanoAddressProjectedNfts` function you wrote

packages/engine/paima-sm/src/types.ts
```diff
@@ -262,6 +288,20 @@ export type ChainDataExtensionCardanoDelegation = ChainDataExtensionBase &
...
+  startSlot: Type.Number(),
+  stopSlot: Type.Optional(Type.Number()),
+  name: Type.String(),
```
SebastienGllmt, line 297: I wonder if we should allow specifying a list of NFT collections you want to filter too. It seems like maybe over-optimizing, but it may cut down on the size of the Carp requests so it may actually make a difference (since if many contracts are using projected NFTs, filtering Carp queries for a specific set of collections may reduce the number of times you have to paginate)

Not a high priority for v1 I guess

packages/engine/paima-sm/src/types.ts
gostkin: this is unused anyway now

packages/engine/paima-sm/src/types.ts
gostkin: I probably don't fully understand the role of `scheduledPrefix`, could you explain pls?

packages/engine/paima-sm/src/types.ts
gostkin: we can do it easily, but i'd change `asset` to `policy id` and `asset name` for better search options

packages/engine/paima-sm/src/types.ts
gostkin: good point, I've fixed the types in carp, that would require new release of client lib so it is updated here too

packages/engine/paima-sm/src/types.ts
SebastienGllmt: `scheduledPrefix` is the event that get scheduled to be sent over to the state machine for the game

For example, you can see entrypoint for the state machine of a game [here](https://github.com/PaimaStudios/paima-game-templates/blob/main/rock-paper-scissors/state-transition/src/stf/v1/index.ts#L28) use a switch-case on the prefix of the message it receives from Paima.

packages/engine/paima-sm/src/types.ts
gostkin: updated

packages/engine/paima-sm/src/types.ts
gostkin: https://github.com/dcSpark/carp/pull/167

packages/engine/paima-rest/src/tsoa/routes.ts
```diff
@@ -0,0 +1,347 @@
+/* tslint:disable */
```
SebastienGllmt, line 1: don't commit this file

packages/engine/paima-funnel/src/cde/cardanoProjectedNFT.ts
```diff
@@ -0,0 +1,67 @@
...
+  url: string,
+  extension: ChainDataExtensionCardanoProjectedNFT,
+  fromAbsoluteSlot: number,
+  toAbsoluteSlot: number,
+  getBlockNumber: (slot: number) => number
+): Promise<ChainDataExtensionDatum[]> {
+  const events = await timeout(
+    query(url, Routes.projectedNftEventsRange, {
+      range: { minSlot: fromAbsoluteSlot, maxSlot: toAbsoluteSlot },
+    }),
```
SebastienGllmt, line 20: Will this handle pagination properly?

gostkin: yes, as long as the range is not too big it will work fine

gostkin: and the request is not split to pages

.gitignore
```diff
@@ -9,4 +9,6 @@ bin
 # logs
 *.log
 
-.idea/
\ No newline at end of file
+.idea/
+
+packages/engine/paima-rest/src/tsoa/routes.ts
```
SebastienGllmt, line 14: We can't add this to gitingore I think because it will break some other systems that require parsing this file. The real solution would be trying to find out why this file sometimes gets formatted by a tool (possibly prettier) despite being specified in the `.prettierignore`. Running `prettier` on the git repo as expected ignores this file, but it's possible `nx` or some other tool is not respecting this somewhere

gostkin: but how git relates to these systems?

gostkin: it's just a file that is not tracked by the repo if it is ignored

gostkin: why would it break anything? otherwise one should always specify all subfolders or do `git rm` on this file which is not very convenient. if we don't need a file in a repo it's better to ignore it imo if it doesn't break things

SebastienGllmt: Many systems and tools look at git for their behavior

SebastienGllmt:
We also need to update the Paima docs with information about this CDE as well

.gitignore
gostkin: as discussed on the call i added back this file since it was not tracked by git and removed it from gitignore

gostkin:
> We also need to update the Paima docs with information about this CDE as well

will add, are the docs blocking this PR to be merged?

SebastienGllmt (approved):

PR #261, opened 2023-11-30
export specificWalletLogin

acedward:


packages/paima-sdk/paima-mw-core/src/index.ts
```diff
@@ -74,6 +75,7 @@ export {
   PaimaMiddlewareErrorCode,
   PAIMA_MIDDLEWARE_ERROR_MESSAGES,
   walletToName,
+  specificWalletLogin,
```
SebastienGllmt, line 78: Oh my bad, I didn't realize we already had `userWalletLogin` exposed for this

PR #265, opened 2023-12-07
carp funnel stake delegation cde epoch tracking

ecioppettini:
### Problem

With the current implementation there is no way to figure out if the delegation has any effect, since there has to be an epoch boundary before that happens.

### Changes

This changes the `getCardanoAddressDelegation` to return either:

1. The last delegation event, if we crossed an epoch boundary already.
2. Both the delegation of the current epoch, plus the previous one.

The current epoch is now stored in the new `cardano_last_epoch` table.

Also `cde_cardano_pool_delegation` now stores the epoch. The epoch is part of the primary key, so only the last event for a certain epoch is in the db. But also only a maximum of 2 rows per address are stored. If the current epoch is greater than both of the epochs in those rows, the earliest one of those is actually ignored in the query, but it's not removed until another delegation tx is indexed for that address.

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
@@ -8,11 +8,10 @@ import {
   Network,
   timeout,
 } from '@paima/utils';
-import type { ChainDataExtensionCardanoProjectedNFT } from '@paima/sm';
+import type { ChainDataExtensionCardanoProjectedNFT, InternalEvent } from '@paima/sm';
 import {
   type ChainData,
   type ChainDataExtension,
-  type ChainDataExtensionCardanoDelegation,
   type ChainDataExtensionDatum,
   type PresyncChainData,
 } from '@paima/sm';
@@ -25,7 +24,7 @@ import getCdePoolData from '../../cde/cardanoPool.js';
 import getCdeProjectedNFTData from '../../cde/cardanoProjectedNFT.js';
 import { query } from '@dcspark/carp-client/client/src/index';
 import { Routes } from '@dcspark/carp-client/shared/routes';
-import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils/src/constants';
+import { FUNNEL_PRESYNC_FINISHED, InternalEventType } from '@paima/utils/src/constants';
 import { CarpFunnelCacheEntry } from '../FunnelCache.js';
 
 const delayForWaitingForFinalityLoop = 1000;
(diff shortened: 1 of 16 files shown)


packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -182,11 +183,30 @@ export async function internalGetAllOwnedErc6551Accounts(
...
   }
 
-  return results[0].pool;
+  const currentEpoch = await getCardanoEpoch.run(undefined, readonlyDBConn);
+
+  if (currentEpoch.length === 0) {
+    throw new Error('Current epoch table not initialized');
+  }
+
+  if (currentEpoch[0].epoch === results[results.length - 1].epoch) {
```
SebastienGllmt, line 198: Are you sure this `if` condition shouldn't be flipped?

I think probably we don't even need this if/else. We can use the same code and then just use `Math.min(0, results.length - 1)` for the index

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -122,11 +156,34 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
...
+        const prevEpoch = this.cache.getState().epoch;
+
+        if (!prevEpoch || epoch !== prevEpoch) {
+          data.internalEvents?.push({
+            type: InternalEventType.CardanoBestEpoch,
+            epoch: epoch,
+          });
+        }
+
+        this.cache.updateEpoch(epoch);
```
SebastienGllmt, line 183: do we want to update the cache like this for every iteration of this loop? (genuine question)

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
ecioppettini: Flipped? Like negating the condition?

This is checking if one of the entries is for the current epoch, in which case all the entries are returned (there can be 2 max), so in this branch `events` is the entire list.

In the other branch instead there is only one result max, because both have to be `< currentEpoch`.

We could simplify the logic by just always returning all of the entries though. I just thought it was leaking implementation details.

Also this `Math.min(0, results.length - 1)` would be always 0, since there is already an early return for the case when there are no delegations, so I'm not sure what you mean.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
ecioppettini: Not really, but afaik the cache it's just another object in memory, so I'm not sure why that would be a problem. Otherwise I could change it to a local variable and then just update it at the end.

Or I can wrap the update in the `epoch !== prevEpoch` condition.

SebastienGllmt: yeah it's less about performance and more about understanding when the cache is meant to be updated. If we only care about updating it when `epoch !== prevEpoch` we should put it in the same condition so it's clear when this cache is mean to be updated at a glance

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
SebastienGllmt: Okay, I think I misunderstood the code originally

We should have a comment block that explains what is going on here 

something like
```js
/**
 * If the most recent delegation is the current epoch, we need to return the list of recent delegations so the app can know what delegation the user had beforehand since delegations only matters once they cross an epoch boundary
 * If the most recent delegation isn't from the current epoch, we know it's the one that is active now
*/
```

either that, or we explain this in the Paima Engine docs and then put a link to the docs as the comment (to avoid the same explanation being in both places)

ecioppettini: I did put an explanation of the result at https://github.com/PaimaStudios/paima-engine-docs/pull/19 

So I could link to that. Although I would be a bit concerned about the domain/url changing if I link to the public version, so I feel like it's better to just replicate it here, maybe a shorter version.

SebastienGllmt (approved):

PR #272, opened 2023-12-15
Feature/delegation

acedward:
# DRAFT.


packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,302 @@
...
+
+type ParsedDelegateWalletCommand =
+  | {
+      command: 'delegate';
+      args: { from: string; to: string; from_signature: string; to_signature: string };
+    }
+  | {
+      command: 'migrate';
+      args: { from: string; to: string; from_signature: string; to_signature: string };
+    }
```
SebastienGllmt, line 31: Initially I had `delegate` and `migrate` as different things. I think it still makes sense conceptually to separate these, but in practice it may be a little tricky so if you want to only have `delegate` and no `migrate` that's fine I think.

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,302 @@
...
+        await updateDelegateFrom.run(
+          {
+            new_from: masterAddress.id,
+            old_from: delegation.from_id,
+            old_to: delegation.set_id,
+          },
+          this.DBConn
+        );
+      }
+    }
```
acedward, line 204: Not supported without merge.

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,302 @@
...
+    // If toAddress had "to" delegations, delete them.
+    // If there toAddress had delegations, update them.
+    const [toAddressHasTo] = await getDelegationsTo.run(
+      { set_id: toAddress.address.id },
+      this.DBConn
+    );
+    if (toAddressHasTo) {
+      await deleteDelegationTo.run({ set_id: toAddress.address.id }, this.DBConn);
+    }
+  }
```
acedward, line 217: Not supported

packages/engine/paima-sm/src/delegate-wallet.ts
acedward: They are implemented, but I can remove them if you don't want them.

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,265 @@
...
+    await newAddress.run({ address: address }, this.DBConn);
+    const [createdAddress] = await getAddressFromAddress.run({ address }, this.DBConn);
+    return { address: createdAddress, isNew: true };
+  }
+
+  // Swap:
+  // addressA gains addressB ID
+  // addressB is assigned a new ID
+  private async swap(addressA: string, addressB: string): Promise<void> {
+    await deleteAddress.run({ address: addressA }, this.DBConn);
```
SebastienGllmt, line 118: why do we delete the address here?

packages/paima-sdk/paima-mw-core/src/wallet-connect/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+    const blockHeightTX = result.blockHeight;
+    await awaitBlock(blockHeightTX);
+    return { success: true, result };
+  } catch (err) {
+    return errorFxn(PaimaMiddlewareErrorCode.ERROR_POSTING_TO_CHAIN, err);
+  }
+}
+
+export type WalletConnectHelperTypes = 'EVM' | 'Cardano' | 'Polkadot' | 'Algorand';
```
SebastienGllmt, line 102: Why do we need this? We already have the `AddressType` enum

packages/paima-sdk/paima-mw-core/src/wallet-connect/index.ts
```diff
@@ -0,0 +1,182 @@
...
+        };
+        break;
+      case 'Polkadot':
+        loginInfo = { mode: WalletMode.Polkadot };
+        break;
+      case 'Algorand':
+        loginInfo = { mode: WalletMode.Algorand };
+        break;
+      default:
+        throw new Error('No supported wallet type.');
```
SebastienGllmt, line 168: It's usually better to use `assertNever` in default cases like this

packages/paima-sdk/paima-mw-core/src/wallet-connect/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    walletName: string,
+    subMessage: string
+  ): Promise<{
+    message: string;
+    signedMessage: string;
+    walletAddress: string;
+  }> {
+    let loginInfo: LoginInfo;
+    switch (walletType) {
+      case 'EVM':
```
SebastienGllmt, line 143: Don't use magic strings (here any many other places)

packages/paima-sdk/paima-mw-core/src/wallet-connect/index.ts
```diff
@@ -0,0 +1,182 @@
...
+      WalletConnectHelper.SEP
+    }${subMessage.toLocaleLowerCase()}${WalletConnectHelper.SEP}${ENV.CONTRACT_ADDRESS}`;
+  }
+
+  private async signWithExternalWallet(
+    walletType: WalletConnectHelperTypes,
+    message: string
+  ): Promise<string> {
+    switch (walletType) {
+      case 'EVM':
```
SebastienGllmt, line 119: Don't use magic strings (here any many other places)

packages/node-sdk/paima-db/src/delegate-wallet.ts
```diff
@@ -0,0 +1,71 @@
...
+  _address: string,
+  DBConn: PoolClient
+): Promise<{
+  from: IGetDelegationsFromWithAddressResult[];
+  to: IGetDelegationsToWithAddressResult[];
+}> {
+  const address = _address.toLocaleLowerCase();
+  const [addressResult] = await getAddressFromAddress.run({ address }, DBConn);
+  const from = await getDelegationsFromWithAddress.run({ from_id: addressResult.id }, DBConn);
+  const to = await getDelegationsToWithAddress.run({ to_id: addressResult.id }, DBConn);
```
SebastienGllmt, line 69: We really should try and batch SQL queries where we can otherwise we pay a round-trip latency (JS -> PSQL -> JS) every time.

If we really can't avoid it in some cases and the queries are truly independent, we can at least to them in parallel with Promise.all

acedward: Fixed as single query

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -100,53 +103,70 @@ export class DelegateWallet {
...
       },
       this.DBConn
     );
-    await newAddress.run({ address: addressB }, this.DBConn);
+    await newAddress.run({ address: addressA.address.address }, this.DBConn);
+
+    return {
+      newA: (await this.getAddress(addressA.address.address)) as AddressExists,
+      newB: (await this.getAddress(addressB.address)) as AddressExists,
+    };
```
SebastienGllmt, line 140: prefer `Promise.all` if we really can't do a single query

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -157,21 +177,25 @@ export class DelegateWallet {
...
+      fromAddress = newA;
+      toAddress = newB;
+    } else {
+      // Case  2:
+      // Do not swap progress.
+      // Create new addresses
+      if (!fromAddress.exists)
+        fromAddress = { address: await this.createAddress(fromAddress.address), exists: true };
+      if (!toAddress.exists)
+        toAddress = { address: await this.createAddress(toAddress.address), exists: true };
```
SebastienGllmt, line 198: Prefer `Promise.all` if we really can't do a single query

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -193,17 +217,22 @@ export class DelegateWallet {
       this.DBConn
     );
 
+    // TODO this is clears the entire cache. We can only clear necessary elements.
```
SebastienGllmt, line 220: TODO

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -193,17 +217,22 @@ export class DelegateWallet {
...
   // Migrate.
   // To gains From ID, and From is assigned a new ID.
   private async cmdMigrate(from: string, to: string): Promise<void> {
-    const [fromAddress] = await getAddressFromAddress.run({ address: from }, this.DBConn);
-    if (!fromAddress) throw new Error('Invalid Address');
-
-    const toAddress = await this.getOrCreateNewAddress(to);
-    await this.swap(fromAddress.address, toAddress.address.address);
+    let fromAddress = await this.getAddress(from);
+    let toAddress = await this.getAddress(to);
```
SebastienGllmt, line 228: Do we need to use `let` here? Also, again with Promise.all

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -213,6 +242,8 @@ export class DelegateWallet {
     const [toAddress] = await getAddressFromAddress.run({ address: to }, this.DBConn);
     if (!toAddress) throw new Error('Invalid Address');
     await deleteDelegationsFrom.run({ from_id: toAddress.id }, this.DBConn);
+
+    // TODO this is clears the entire cache. We can only clear necessary elements.
```
SebastienGllmt, line 246: TODO

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -193,17 +217,22 @@ export class DelegateWallet {
...
-    await this.swap(fromAddress.address, toAddress.address.address);
+    let fromAddress = await this.getAddress(from);
+    let toAddress = await this.getAddress(to);
+    await this.validateDelegate(fromAddress, toAddress);
+    if (!fromAddress.exists && toAddress.exists) {
+      await this.swap(toAddress, fromAddress);
+    } else {
+      throw new Error('Cannot migrate');
+    }
+    // TODO this is clears the entire cache. We can only clear necessary elements.
```
SebastienGllmt, line 235: TODO

packages/engine/paima-sm/src/delegate-wallet.ts
acedward: Queries merged

packages/node-sdk/paima-db/migrations/up.sql
```diff
@@ -98,6 +98,59 @@ CREATE TABLE delegations (
  PRIMARY KEY (from_id, to_id)
 );
 
+-- Create a function to notify any change in address or delegate tables
+create or replace function public.notify_wallet_connect()
```
SebastienGllmt, line 102: Although I think this is not bad to avoid blocking on this, but I can't image this scaling well if we need to cache more things in the future like the achievement progress of users since this is a lot of code and complexity per cache invalidation logic

acedward: This a generic function, it transforms any trigger into a nice json, with table / method / data 
```
create or replace function public.notify_wallet_connect()
returns trigger
  language plpgsql
as $function$ 
begin
   perform pg_notify('wallet_connect_change', payload);
    return null;
end;
$function$
;
``` 

Is enough, but does not differentiate between table/method or give info about changed data

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,308 @@
...
+      return; // both are new.
+    }
+    if (fromAddress.address.id === toAddress.address.id) {
+      throw new Error('Cannot delegate to itself');
+    }
+    if (fromAddress.exists && toAddress.exists) {
+      throw new Error('Both A and B have progress. Cannot merge.');
+    }
+
+    await Promise.all([
```
SebastienGllmt, line 154: Probably the code will be a lot cleaner if instead of checking the error message inside the promise, you return a tuple of 3 value and then throw the errors afterwards

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,308 @@
...
+      {
+        new_address: nonExistingAddress.address,
+        old_address: existingAddress.address.address,
+      },
+      this.DBConn
+    );
+    await newAddress.run({ address: existingAddress.address.address }, this.DBConn);
+
+    return await Promise.all([
+      this.getAddress(existingAddress.address.address) as Promise<AddressExists>,
```
SebastienGllmt, line 135: Do you really need the `as` conversions here?

packages/paima-sdk/paima-mw-core/src/wallet-connect/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  public buildMessageToSign(subMessage: string): string {
+    return `${WalletConnectHelper.DELEGATE_WALLET_PREFIX}${
+      WalletConnectHelper.SEP
+    }${subMessage.toLocaleLowerCase()}${WalletConnectHelper.SEP}${ENV.CONTRACT_ADDRESS}`;
+  }
+
+  private async signWithExternalWallet(walletType: AddressType, message: string): Promise<string> {
+    switch (walletType) {
+      case AddressType.EVM:
+        return (await EvmInjectedConnector.instance().getProvider()?.signMessage(message)) || '';
```
SebastienGllmt, line 116: Is returning `''` here really valid? It looks to me like it would cause things to break, so it's better to fail fast than to silently return the wrong result and fail much later imo

SebastienGllmt:
We also need to update the Paima docs with information about how to use this system as well

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -0,0 +1,307 @@
...
+      CryptoManager.Polkadot(),
+    ];
+    for (const wallet of wallets) {
+      try {
+        if (await wallet.verifySignature(walletAddress, message, signature)) {
+          return true;
+        }
+      } catch {
+        // do nothing.
+      }
```
SebastienGllmt, line 105: This is going to run really slowly (esp. if we add more currencies). Probably we want to detect which algorithm to use in some other way (either by checking which address submitted this or through an extra field)

acedward: By address (would this work?)

eth: `'0x[a-fA-F0-9]{40}'` 
cardano `addr1[a-z0-9]+` 
polkadot `[1-9A-HJ-NP-Za-km-z]{47,48}` https://namespaces.chainagnostic.org/polkadot/caip10
algorand ` [A-Z2-7]{58}` 

We would have to add 2 params to wallet delegation 
`&wd|from|to|from_signature|to|signature|from_type|to_type`

What path do we take?

SebastienGllmt: I don't think we need to add anything to the concise encoding:
1. If the user uses the batcher, we already know the address since the batch is of the form `userAddress/userSignature/input/millisecondTimestamp`
2. If the user self-sequences, we know the address since they're the ones who submitted the transaction

SebastienGllmt:
This PR also needs to be rebased

packages/engine/paima-sm/src/delegate-wallet.ts
```diff
@@ -82,26 +82,52 @@ export class DelegateWallet {
...
   }
 
+  private validateSender(to: string, from: string, realAddress: string): void {
+    if (to === from) throw new Error('Cannot delegate self');
+    const isSelf = to === realAddress || from === realAddress;
+    if (!isSelf) throw new Error('Either to or from must bt the sender address');
+  }
+
+  // Regex must match all possible wallets for the network.
+  private static readonly WALLET_VALIDATORS: { regex: RegExp; manager: () => IVerify }[] = [
```
SebastienGllmt, line 92: We have `verifyAddress` as part of the `@paima/crypto` package for this

SebastienGllmt (approved):

Issue #264, opened 2023-12-05
Cardano delayed asset CDE


SebastienGllmt:
Currently we have the projected NFT system to allow Paima to asset that a user holds a specific asset right now

But for some use-cases, we don't care if they hold something right now (since assert this involves extra steps and txs fees) and just knowing they hold it recently is good enough

Additionally, this kind of delayed check is also useful for projects that want to check the user's balance of a non-NFT token

We should provide this as an alternative to the projected NFT CDE. The way this would work is that Paima tracks all the UTXO entries for the specified asset

SebastienGllmt:
Implemented by #275

PR #275, opened 2023-12-27
delayed asset cde

ecioppettini:
Implements #264 

I'm currently uncertain about what's the proper api for this.

Currently this has to be configured like this:

```
extensions:
  - name: "assets1"
    type: cardano-delayed-nft
    assets:
      - "asset1c43p68zwjezc7f6w4w9qkhkwv9ppwz0f7c3amw"
    startSlot: 22505578
```

Which implies that the asset is an nft, but currently the utils function is used like this.

```
  const utxos = await getCardanoAssetUtxos(pool, "8200581cd3d8e0b241c10cc0e70300b91d79536af390c6c7b1c919687afc3e2f", "asset1c43p68zwjezc7f6w4w9qkhkwv9ppwz0f7c3amw");
```

But I'm not sure if returning the actual utxo entries makes sense.

Since the issue also mentions this:

> Additionally, this kind of delayed check is also useful for projects that want to check the user's balance of a non-NFT token

I'm not really sure what's more convenient. Maybe the cde should be named `cardano-delayed-token` or something like that?


packages/engine/paima-funnel/package.json
```diff
@@ -18,6 +18,6 @@
   },
   "dependencies": {
     "assert-never": "^1.2.1",
-    "@dcspark/carp-client": "^2.3.0"
+    "@dcspark/carp-client": "file://../../../tmp-carp-client"
```
SebastienGllmt, line 21: Just a reminder we'll have to update this before merging

tmp-carp-client/client/src/index.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
```
SebastienGllmt, line 1: I don't think you meant to check this folder in

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -190,3 +191,16 @@ export async function internalGetCardanoAddressDelegation(
 
   return results[0].pool;
 }
+
+export async function internalGetCardanoAssetUtxos(
+  readonlyDBConn: Pool,
+  address: string,
+  cip14Fingerprint: string
```
SebastienGllmt, line 198: Can we add an option to query by policy ID as well? Tracking a specific asset is useful for fungible tokens, and policy IDs useful for NFTs

We could make these both the same CDE (called something like delayed asset or delayed UTXO), or if it's easier to split them we can have the asset one be delayed token and the policy ID be delayed NFT

tmp-carp-client/client/src/index.js
ecioppettini: Well, I did it in case someone wanted to checkout this branch and run it. Although currently there is no CI, it won't build without the dependency. Maybe it _would_ be better to put the code in a repo and point npm to it though, so I don't have to remember to untrack this later.

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
ecioppettini: I guess then we need to also be able to ask carp for the utxos by policy id, then. Otherwise you would have to specify in the cde config all of the fingerprints for that policy id. This is doable though, but I need to add it to carp first.

SebastienGllmt: Yeah it requires a change in Carp, but it's just a new query on the same data so it should be a fairly small change

ecioppettini: I think it's easier to keep them together, but I'm not sure about the API. Right now it can be configured like this:

```
extensions:
  - name: "assets1"
    type: cardano-delayed-nft
    fingerprints:
      - "asset1c43p68zwjezc7f6w4w9qkhkwv9ppwz0f7c3amw"
    policyIds:
      - "919d4c2c9455016289341b1a14dedf697687af31751170d56a31466e"
    startSlot: 22505578
```

Where both `fingerprings` and `policyIds` are optional (so one or the other can be used).

And I split the utils in `getCardanoAssetUtxosByFingerprint` and `getCardanoAssetUtxosByPolicyId`.

I could potentially rename the CDE to delayed utxo, but I'm not sure if mixing and matching fingerprints with policyids is a good idea or not. It may be better to have two different configs, and just use the same code for both?

packages/node-sdk/paima-utils-backend/src/cde-access.ts
```diff
@@ -167,3 +168,24 @@ export async function getCardanoAddressDelegation(
...
+  );
+}
+
+export async function getCardanoAssetUtxosByPolicyId(
+  readonlyDBConn: Pool,
+  address: string,
+  policyId: string
+): Promise<{ txId: string; outputIndex: number; amount: string }[]> {
+  return await internalGetCardanoAssetUtxos(readonlyDBConn, address, 'policy_id', policyId);
+}
```
SebastienGllmt, line 191: Probably for both of these we need to include the assetName in the return type - otherwise there is no way for the game to know which asset specifically this is referring to

Recall that going from `<policy_id, asset_name>` → fingerprint is one-direction (so unless you also get the policy+asset back, games would have to also include a inverse map which isn't very convenient)

packages/engine/paima-sm/src/types.ts
```diff
@@ -145,6 +161,7 @@ export enum CdeEntryTypeName {
   ERC721 = 'erc721',
   ERC6551Registry = 'erc6551-registry',
   CardanoDelegation = 'cardano-stake-delegation',
+  CardanoDelayedNft = 'cardano-delayed-nft',
```
SebastienGllmt, line 164: I think we have to change this name. I don't really mind if it's one CDE or two CDEs (if we change our minds later, we can just change the code). It should be something more general than `nft` like "token" or "asset" though

packages/node-sdk/paima-utils-backend/src/cde-access.ts
ecioppettini: Hmmm, yeah. It's actually included in the result... I just forgot to update the type. And also forgot that typescript worked this way.

Currently the result is something like this:

```
utxo {
  txId: '2a35d3e5e55dcf2538826c66989baccfbded226fd70992155f841cd8866b1d6e',
  outputIndex: 0,
  amount: '16502879677629',
  policyId: '919d4c2c9455016289341b1a14dedf697687af31751170d56a31466e',
  assetName: '7449555344'
}
```

Although I guess

packages/engine/paima-sm/src/cde-cardano-delayed-asset.ts
```diff
@@ -0,0 +1,36 @@
...
+  const cdeId = cdeDatum.cdeId;
+  const address = cdeDatum.payload.address;
+  const amount = cdeDatum.payload.amount;
+  const tx_id = cdeDatum.payload.txId;
+  const output_index = cdeDatum.payload.outputIndex;
+  const cip14_fingerprint = cdeDatum.payload.cip14Fingerprint;
+  const policy_id = cdeDatum.payload.policyId;
+  const asset_name = cdeDatum.payload.assetName;
+
+  if (amount) {
```
SebastienGllmt, line 17: Maybe a comment to explain this

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -229,3 +235,22 @@ export async function internalGetCardanoProjectedNft(
 
   return results;
 }
+
+export async function internalGetCardanoAssetUtxos(
+  readonlyDBConn: Pool,
+  address: string,
+  filterBy: string,
```
SebastienGllmt, line 242: I think you can replace this by `filterBy: keyof ICdeCardanoAssetUtxosByAddressResult` or just `filterBy: 'policy_id' | 'cip14_fingerprint'`

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -229,3 +236,22 @@ export async function internalGetCardanoProjectedNft(
...
+
+export async function internalGetCardanoAssetUtxos(
+  readonlyDBConn: Pool,
+  address: string,
+  filterBy: keyof ICdeCardanoAssetUtxosByAddressParams,
+  value: string
+): Promise<CardanoAssetUtxo[]> {
+  const arg = { address, [filterBy]: value };
+
+  const results = await cdeCardanoAssetUtxosByAddress.run(arg, readonlyDBConn);
```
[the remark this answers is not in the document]
SebastienGllmt: Hmm, you're probably right

SebastienGllmt (approved):

PR #277, opened 2024-01-09
Improve Solidity usability

SebastienGllmt:
Right now deploying and interacting with the Solidity contracts for Paima is a pain. The old setup was like this:
1. Manually extract the contracts from Paima using the `./paima-engine contracts` CLI
2. Modify the contracts as needed
3. Setup your own tool to deploy them (or use a custom CLI we provided that was extremely buggy)

The new option is like this:
1. Improve any contract you need from NPM directly into your game (only overriding things if necessary)
2. Run a simple deploy command to deploy the contracts to a custom chain or production
3. Interact with these contracts using hardhat (either `hardhat-interact` or custom Paima tasks)

Some tasks off the top of my head that still need to be done
- [x] properly bundle the Solidity contracts with the Paima executable for the `paima contracts` command
- [x] decide if/how we want to include `typechain-types` in the NPM package
- [x] fix prettier complaining about the `evm-contracts` folder (I don't know why it doesn't like this despite being fine with the old code)
- [ ] port any tests we want to keep from the old folders to the new `evm-contracts` folder
- [ ] improve documentation of the contracts (that is to say, inline docs inside the `.sol` files
- [ ] review the contracts (they were written quite a while ago, so probably best to review them  to make sure they still follow best practices for Solidity)
- [x] delete the old `paima-l2-contract` and `nft` folders
- [x] publish the `evm-contracts` package to NPM

Related PRs that are ongoing
- update the Paima docs to explain the new process
- update Paima game templates to the new process


packages/contracts/evm-contracts/contracts/Proxy/NativeProxy.sol
```diff
@@ -0,0 +1,71 @@
...
+pragma solidity ^0.8.7;
+
+import "../ERC1967.sol";
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+/// @title Proxy
+/// @dev Proxy contract mostly based on https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Proxy.sol
+contract NativeProxy is ERC1967 {
```
matejos, line 12: This is a proxy for NativeNftSale contract right? I would propose changing the name to NativeNftSaleProxy, so it's more explanatory and consistent with the Proxy naming

packages/contracts/evm-contracts/contracts/Proxy/Proxy.sol
```diff
@@ -0,0 +1,73 @@
...
+pragma solidity ^0.8.7;
+
+import "../ERC1967.sol";
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+/// @title Proxy
+/// @dev Proxy contract mostly based on https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Proxy.sol
+contract Proxy is ERC1967 {
```
matejos, line 12: This is a proxy for Erc20NftSale contract right? I would propose changing the name to Erc20NftSaleProxy, so it's more explanatory and consistent with the Proxy naming

packages/contracts/evm-contracts/contracts/State.sol
```diff
@@ -0,0 +1,24 @@
...
+
+    bool public initialized;
+
+    // NFT address
+    address public nftAddress;
+
+    // Array of deposited currencies
+    ERC20[] public depositedCurrencies;
+
+    // mapping from token address to amount
```
matejos, line 19: That's not true, it's a mapping from token address to boolean representing if this token has been deposited

packages/contracts/evm-contracts/contracts/State.sol
```diff
@@ -0,0 +1,24 @@
...
+    mapping(ERC20 => bool) public depositedCurrenciesMap;
+
+    // Array of supported currencies
```
matejos, line 22: Comments like these feel really unnecessary 😅

PR #282, opened 2024-01-12
add cml dependency

ecioppettini:
Not necessarily a PR to merge, mostly to isolate a problem that I don't know how to fix, although we could remove the test code and merge it if it works.

The code in the block funnel is there only to test this without needing to setup a cde, just running the engine should be enough.

The problem:

*version*: ^3.0.1

I get this error when bootstrapping:

```
LinkError: WebAssembly.Instance(): Import #4 module="__wbindgen_placeholder__" function="__wbg_new_1d9a920c6bfc44a8" error: function import requires a callable
```

*version*: 4.0.2 (latest)

This version was giving a similar error to me yesterday, but now it actually manages to run partially, so there may be other issue around. The problem is that when the binding code uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN it gets undefined (for example, when trying to get the amount of a multiasset from an asset name).

I think this may be because the class does not get extended correctly. The code in `engineCorePacked.js` has something like this

```ts
var BigInt4 = class _BigInt {
      static __wrap(ptr) {
        const obj2 = Object.create(_BigInt.prototype);
     ...
```

Which does not seem correct? (The original uses `BigInt` in both places).

Ideally we would use this version, but I tried both just in case it was only an issue with `BigInt` support.

### DIFF
--- a/packages/engine/paima-funnel/package.json
+++ b/packages/engine/paima-funnel/package.json
@@ -19,6 +19,6 @@
   "dependencies": {
     "assert-never": "^1.2.1",
     "@dcspark/carp-client": "^2.3.1",
-    "@dcspark/cardano-multiplatform-lib-nodejs": "^3.0.1"
+    "@dcspark/cardano-multiplatform-lib-nodejs": "^5.1.0"
   }
 }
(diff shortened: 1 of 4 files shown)


ecioppettini:
Closing this since it was mostly supposed to be a shortcut to reproduce the problem/get the error.

PR #286, opened 2024-01-16
EVM contracts fixes and documentation

matejos:


packages/contracts/evm-contracts/contracts/README.md
```diff
@@ -46,4 +46,8 @@ Facilitating monetization
 
 {{Proxy}}
```
SebastienGllmt, line 47: I think this needs to be deleted (since it was replaced with `Erc20NftSaleProxy` below in this file)

SebastienGllmt: Similarly, `NativeProxy` should be removed since we added `NativeNftSaleProxy` instead

packages/contracts/evm-contracts/contracts/AnnotatedMintNft.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint256 _tokenId
+    ) external onlyExistingTokenId(_tokenId) onlyTokenOwner(_tokenId) {
+        totalSupply--;
+        _burn(_tokenId);
+    }
+
+    /// @dev Adds `_minter` to the mapping of allowed `minters` of this NFT.
+    /// Callable only by the contract owner.
+    /// Emits the `SetMinter` event.
+    function setMinter(address _minter) external onlyOwner {
```
SebastienGllmt, line 118: I realized today that the `AnnotatedMintNft` cannot be used by default (the `canMint` modifier will throw`), so to use it you NEED to set at least 1 minter (or be the contract owner). My guess is that the reason for this is to disable minting unless it's done through the `NativeNftSale` contract (and it supports multiple minters in case you want to provide multiple ways to mint the NFT)

Probably this design should be mentioned in the docs

matejos: Added a contract-level comment about this.

SebastienGllmt (approved):

PR #300, opened 2024-02-02
parallel evm chain funnel sync + move multiple network configurations to a file

ecioppettini:
# About

This PR introduces a new funnel type, called evm funnel. This funnel gets instantiated for each additional evm network that needs to be synced. For the presync it works similar to the block funnel, but for the sync phase it does act differently. Here we get first the data from the block funnel, and then these blocks are paired with the CDE events from the additional evm chain.

The pairing works by using timestamps, so a block from the new chain gets assigned to the first block in the main chain with a timestamp greater or equal than its own. This is done in order to provide determinism.

## Config changes

The network configuration is moved to a file, which should be named `config.$NETWORK.yml` (not `.yaml`). This is because otherwise difficult to namespace variable names for each different network.

```yaml
# config.localhost.yml (config.$NETWORK.yml)
Hardhat:
  type: evm-main
  chainUri: 'http://localhost:8545'
  chainId: 31337
  chainCurrencyName: 'Test Hardhat Tokens'
  chainCurrencySymbol: 'TEST'
  chainCurrencyDecimals: 18
  blockTime: 2
  paimaL2ContractAddress: '0x5FbDB2315678afecb367f032d93F642f64180aa3'

Arbitrum:
  type: evm-other
  chainUri: 'https://arb1.arbitrum.io/rpc'
  chainId: 42161
  chainCurrencyName: 'ETH'
  chainCurrencySymbol: 'ETH'
  chainCurrencyDecimals: 18
  blockTime: 2
  # this regulates the amount of blocks fetched during the sync in a single round. It helps to avoid rate-limiting.
  funnelBlockGroupSize: 20
```
*NOTE:* The *funnelBlockGroupSize* in the example is what I had to set in order to not get rate limited with arbitrum  

And the extensions have to be specified like:

```yaml
# extensions.yml
extensions:
  - name: "Bar"
    type: "erc20"
    network: "Hardhat"
    contractAddress: "0x0000000000000000000000000000000000000001"
    startBlockHeight: 0

  - name: "USDC"
    type: "erc20"
    contractAddress: "0xaf88d065e77c8cC2239327C5EDb3A432268e5831"
    startBlockHeight: 178573900
    # This is required so that the funnel can know which extensions should it care about
    network: "Arbitrum"
```

Note that the network field uses the id of the label in the other object.

I think potentially the extensions.yml file should be removed and just moved to the other file under a field, but I'm not sure.

### TODO
- ? Removed unused ENV vars . *Here I'm not sure if the vars are currently used by other stuff, like frontends, that may not want to read a file*
- ? Store the config and check it which fields shouldn't change 
- ? Probably don't need `evm-main` and `evm-other`, since the `paimaL2ContractAddress` can be used for discrimination anyway .
- [x] Instantiation of one funnel per evm network.
- [x] Figure out how to match blocks across networks.

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/block/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/block/funnel.ts
@@ -1,4 +1,4 @@
-import { ENV, EvmConfig, GlobalConfig, Network, doLog, timeout } from '@paima/utils';
+import { ENV, EvmConfig, GlobalConfig, doLog, timeout } from '@paima/utils';
 import type { ChainFunnel, ReadPresyncDataFrom } from '@paima/runtime';
 import type { ChainData, PresyncChainData } from '@paima/sm';
 import { getBaseChainDataMulti, getBaseChainDataSingle } from '../../reading.js';
@@ -9,6 +9,7 @@ import type { FunnelSharedData } from '../BaseFunnel.js';
 import { RpcCacheEntry, RpcRequestState } from '../FunnelCache.js';
 import type { PoolClient } from 'pg';
 import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
 
 const GET_BLOCK_NUMBER_TIMEOUT = 5000;
 
@@ -88,7 +89,8 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
           this.sharedData.web3,
           this.sharedData.paimaL2Contract,
           blockNumber,
-          this.dbTx
+          this.dbTx,
+          this.chainName
         ),
         getUngroupedCdeData(
           this.sharedData.web3,
@@ -124,7 +126,8 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
           this.sharedData.paimaL2Contract,
           fromBlock,
           toBlock,
-          this.dbTx
+          this.dbTx,
+          this.chainName
         ),
         getUngroupedCdeData(
           this.sharedData.web3,
@@ -133,7 +136,13 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
           toBlock
         ),
       ]);
-      const cdeData = groupCdeData(Network.CARDANO, fromBlock, toBlock, ungroupedCdeData);
+      const cdeData = groupCdeData(
+        this.chainName,
+        ConfigNetworkType.EVM,
+        fromBlock,
+        toBlock,
+        ungroupedCdeData
+      );
       return composeChainData(baseChainData, cdeData);
     } catch (err) {
       doLog(`[funnel] at ${fromBlock}-${toBlock} caught ${err}`);
@@ -143,25 +152,25 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
 
   public override async readPresyncData(
     args: ReadPresyncDataFrom
-  ): Promise<{ [network: number]: PresyncChainData[] | typeof FUNNEL_PRESYNC_FINISHED }> {
-    let arg = args.find(arg => arg.network == Network.EVM);
+  ): Promise<{ [network: string]: PresyncChainData[] | typeof FUNNEL_PRESYNC_FINISHED }> {
+    let arg = args.find(arg => arg.network == this.chainName);
 
     if (!arg) {
-      return [];
+      return {};
     }
 
     let fromBlock = arg.from;
     let toBlock = arg.to;
 
     if (fromBlock >= ENV.START_BLOCKHEIGHT) {
-      return { [Network.EVM]: FUNNEL_PRESYNC_FINISHED };
+      return { [this.chainName]: FUNNEL_PRESYNC_FINISHED };
     }
 
     try {
       toBlock = Math.min(toBlock, ENV.START_BLOCKHEIGHT);
       fromBlock = Math.max(fromBlock, 0);
       if (fromBlock > toBlock) {
-        return [];
+        return {};
       }
 
       const ungroupedCdeData = await getUngroupedCdeData(
@@ -170,7 +179,15 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
         fromBlock,
         toBlock
       );
-      return { [Network.EVM]: groupCdeData(Network.EVM, fromBlock, toBlock, ungroupedCdeData) };
+      return {
+        [this.chainName]: groupCdeData(
+          this.chainName,
+          ConfigNetworkType.EVM,
+          fromBlock,
+          toBlock,
+          ungroupedCdeData
+        ),
+      };
     } catch (err) {
       doLog(`[paima-funnel::readPresyncData] Exception occurred while reading blocks: ${err}`);
       throw err;
--- a/packages/engine/paima-funnel/src/funnels/block/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/block/funnel.ts
@@ -1,4 +1,4 @@
-import { ENV, Network, doLog, timeout } from '@paima/utils';
+import { ENV, EvmConfig, GlobalConfig, Network, doLog, timeout } from '@paima/utils';
 import type { ChainFunnel, ReadPresyncDataFrom } from '@paima/runtime';
 import type { ChainData, PresyncChainData } from '@paima/sm';
 import { getBaseChainDataMulti, getBaseChainDataSingle } from '../../reading.js';
@@ -13,12 +13,22 @@ import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
 const GET_BLOCK_NUMBER_TIMEOUT = 5000;
 
 export class BlockFunnel extends BaseFunnel implements ChainFunnel {
-  protected constructor(sharedData: FunnelSharedData, dbTx: PoolClient) {
+  config: EvmConfig;
+  chainName: string;
+
+  protected constructor(
+    sharedData: FunnelSharedData,
+    dbTx: PoolClient,
+    config: EvmConfig,
+    chainName: string
+  ) {
     super(sharedData, dbTx);
     // TODO: replace once TS5 decorators are better supported
     this.readData.bind(this);
     this.readPresyncData.bind(this);
     this.getDbTx.bind(this);
+    this.config = config;
+    this.chainName = chainName;
   }
 
   public override async readData(blockHeight: number): Promise<ChainData[]> {
@@ -32,10 +42,10 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
     }
 
     if (toBlock === fromBlock) {
-      doLog(`Block funnel ${ENV.CHAIN_ID}: #${toBlock}`);
+      doLog(`Block funnel ${this.config.chainId}: #${toBlock}`);
       return await this.internalReadDataSingle(fromBlock);
     } else {
-      doLog(`Block funnel ${ENV.CHAIN_ID}: #${fromBlock}-${toBlock}`);
+      doLog(`Block funnel ${this.config.chainId}: #${fromBlock}-${toBlock}`);
       return await this.internalReadDataMulti(fromBlock, toBlock);
     }
   }
@@ -53,7 +63,7 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
 
     const latestBlockQueryState = this.sharedData.cacheManager.cacheEntries[
       RpcCacheEntry.SYMBOL
-    ]?.getState(ENV.CHAIN_ID);
+    ]?.getState(this.config.chainId);
     if (latestBlockQueryState?.state !== RpcRequestState.HasResult) {
       throw new Error(`[funnel] latest block cache entry not found`);
     }
@@ -82,7 +92,7 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
         ),
         getUngroupedCdeData(
           this.sharedData.web3,
-          this.sharedData.extensions,
+          this.sharedData.extensions.filter(extension => extension.network === this.chainName),
           blockNumber,
           blockNumber
         ),
@@ -116,7 +126,12 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
           toBlock,
           this.dbTx
         ),
-        getUngroupedCdeData(this.sharedData.web3, this.sharedData.extensions, fromBlock, toBlock),
+        getUngroupedCdeData(
+          this.sharedData.web3,
+          this.sharedData.extensions.filter(extension => extension.network === this.chainName),
+          fromBlock,
+          toBlock
+        ),
       ]);
       const cdeData = groupCdeData(Network.CARDANO, fromBlock, toBlock, ungroupedCdeData);
       return composeChainData(baseChainData, cdeData);
@@ -151,7 +166,7 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
 
       const ungroupedCdeData = await getUngroupedCdeData(
         this.sharedData.web3,
-        this.sharedData.extensions,
+        this.sharedData.extensions.filter(extension => extension.network === this.chainName),
         fromBlock,
         toBlock
       );
@@ -182,8 +197,10 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
       return newEntry;
     })();
 
-    cacheEntry.updateState(ENV.CHAIN_ID, latestBlock);
+    const [chainName, config] = await GlobalConfig.mainEvmConfig();
 
-    return new BlockFunnel(sharedData, dbTx);
+    cacheEntry.updateState(config.chainId, latestBlock);
+
+    return new BlockFunnel(sharedData, dbTx, config, chainName);
   }
 }
--- a/packages/engine/paima-funnel/src/funnels/block/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/block/funnel.ts
@@ -9,7 +9,7 @@ import type { FunnelSharedData } from '../BaseFunnel.js';
 import { RpcCacheEntry, RpcRequestState } from '../FunnelCache.js';
 import type { PoolClient } from 'pg';
 import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
-import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
+import { ConfigNetworkType } from '@paima/utils';
 
 const GET_BLOCK_NUMBER_TIMEOUT = 5000;
 
--- a/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
@@ -263,7 +263,7 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
       ]);
 
       let grouped = groupCdeData(
-        'cardano',
+        this.chainName,
         ConfigNetworkType.CARDANO,
         arg.from,
         arg.to,
--- a/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
@@ -163,7 +163,8 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
       lastTimestamp,
       this.cache,
       this.config.confirmationDepth,
-      this.era
+      this.era,
+      this.chainName
     );
 
     const composed = composeChainData(this.bufferedData, grouped);
@@ -342,7 +343,8 @@ async function readDataInternal(
   lastTimestamp: number,
   cache: CarpFunnelCacheEntry,
   confirmationDepth: number,
-  era: Era
+  era: Era,
+  chainName: string
 ): Promise<PresyncChainData[]> {
   // the lower range is exclusive
   const min = timestampToAbsoluteSlot(era, lastTimestamp, confirmationDepth);
@@ -434,8 +436,7 @@ async function readDataInternal(
   );
 
   let grouped = groupCdeData(
-    // TODO: not really
-    'cardano',
+    chainName,
     ConfigNetworkType.CARDANO,
     data[0].blockNumber,
     data[data.length - 1].blockNumber,
--- a/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
@@ -8,6 +8,7 @@ import {
   GlobalConfig,
   logError,
   timeout,
+  ConfigNetworkType,
 } from '@paima/utils';
 import type { InternalEvent } from '@paima/sm';
 import {
@@ -29,7 +30,6 @@ import { Routes } from '@dcspark/carp-client/shared/routes';
 import { FUNNEL_PRESYNC_FINISHED, InternalEventType } from '@paima/utils';
 import { CarpFunnelCacheEntry } from '../FunnelCache.js';
 import { getCardanoEpoch } from '@paima/db';
-import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
 
 const delayForWaitingForFinalityLoop = 1000;
 
--- a/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/carp/funnel.ts
@@ -7,7 +7,6 @@ import {
   ENV,
   GlobalConfig,
   logError,
-  Network,
   timeout,
 } from '@paima/utils';
 import type { InternalEvent } from '@paima/sm';
@@ -30,6 +29,7 @@ import { Routes } from '@dcspark/carp-client/shared/routes';
 import { FUNNEL_PRESYNC_FINISHED, InternalEventType } from '@paima/utils';
 import { CarpFunnelCacheEntry } from '../FunnelCache.js';
 import { getCardanoEpoch } from '@paima/db';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
 
 const delayForWaitingForFinalityLoop = 1000;
 
@@ -111,7 +111,8 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
     private readonly baseFunnel: ChainFunnel,
     private readonly carpUrl: string,
     private cache: CarpFunnelCacheEntry,
-    private readonly config: CardanoConfig
+    private readonly config: CardanoConfig,
+    private readonly chainName: string
   ) {
     super(sharedData, dbTx);
     // TODO: replace once TS5 decorators are better supported
@@ -200,7 +201,7 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
   public override async readPresyncData(
     args: ReadPresyncDataFrom
   ): Promise<{ [network: number]: PresyncChainData[] | typeof FUNNEL_PRESYNC_FINISHED }> {
-    const arg = args.find(arg => arg.network == Network.CARDANO);
+    const arg = args.find(arg => arg.network == this.chainName);
 
     let basePromise = this.baseFunnel.readPresyncData(args);
 
@@ -260,10 +261,16 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
         basePromise,
       ]);
 
-      let grouped = groupCdeData(Network.CARDANO, arg.from, arg.to, carpEvents);
+      let grouped = groupCdeData(
+        'cardano',
+        ConfigNetworkType.CARDANO,
+        arg.from,
+        arg.to,
+        carpEvents
+      );
 
       if (grouped.length > 0) {
-        data[Network.CARDANO] = grouped;
+        data[this.chainName] = grouped;
       }
 
       return data;
@@ -271,7 +278,7 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
       const data = await basePromise;
 
       if (arg) {
-        data[Network.CARDANO] = FUNNEL_PRESYNC_FINISHED;
+        data[this.chainName] = FUNNEL_PRESYNC_FINISHED;
       }
 
       return data;
@@ -283,7 +290,8 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
     dbTx: PoolClient,
     baseFunnel: ChainFunnel,
     config: CardanoConfig,
-    startingBlockHeight: number
+    startingBlockHeight: number,
+    chainName: string
   ): Promise<CarpFunnel> {
     if (!config.confirmationDepth) {
       throw new Error('[carp-funnel] Missing CARDANO_CONFIRMATION_DEPTH setting.');
@@ -315,7 +323,15 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
       return newEntry;
     })();
 
-    return new CarpFunnel(sharedData, dbTx, baseFunnel, config.carpUrl, await cacheEntry, config);
+    return new CarpFunnel(
+      sharedData,
+      dbTx,
+      baseFunnel,
+      config.carpUrl,
+      await cacheEntry,
+      config,
+      chainName
+    );
   }
 }
 
@@ -418,7 +434,9 @@ async function readDataInternal(
   );
 
   let grouped = groupCdeData(
-    Network.EVM,
+    // TODO: not really
+    'cardano',
+    ConfigNetworkType.CARDANO,
     data[0].blockNumber,
     data[data.length - 1].blockNumber,
     poolEvents.filter(data => data.length > 0)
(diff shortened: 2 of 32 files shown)


packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -0,0 +1,154 @@
...
+  // Parse the YAML content into an object
+  const configObject = YAML.parse(configFileData);
+
+  // Validate the YAML object against the schema
+  const baseConfig = checkOrError(undefined, BaseConfigWithoutDefaults, configObject);
+
+  return baseConfig;
+}
+
+// TODO: copy pasted code
```
SebastienGllmt, line 127: TODO?

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -0,0 +1,154 @@
...
+  blockTime: 4,
+  pollingRate: (blockTime || 4) - 0.1,
+  deployment: '',
+  funnelBlockGroupSize: 100,
+});
+
+export async function loadConfig(): Promise<Static<typeof BaseConfigWithDefaults> | undefined> {
+  let configFileData: string;
+  try {
+    // TODO: would be nice to also read .yaml
```
SebastienGllmt, line 82: TODO?

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -0,0 +1,154 @@
...
+export const FullEvmConfig = EvmConfigSchema;
+
+const evmConfigDefaults = (blockTime: number | undefined) => ({
+  chainUri: '',
+  chainId: 0,
+  chainExplorerUri: '',
+  chainCurrencyName: 'UNKNOWN_CURRENCY_NAME',
+  chainCurrencySymbol: 'NONAME',
+  chainCurrencyDecimals: 0,
+  blockTime: 4,
```
SebastienGllmt, line 73: strange how block time is a parameter, yet here we just set it to 4. I'm not sure why we would even want these defaults instead of just erroring though since other than `chainExplorerUri` and `pollingRate` these are all required (also fyi `deployment` is deprecated)

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -0,0 +1,154 @@
...
+          break;
+        case ConfigNetworkType.CARDANO:
+        default:
+          throw new Error('unknown network type');
+      }
+    }
+
+    return config;
+  } catch (err) {
+    // doLog(`Invalid config file: ${err}`);
```
SebastienGllmt, line 111: is this meant to be commented out?

packages/paima-sdk/paima-mw-core/src/state.ts
```diff
@@ -17,17 +17,18 @@ let gameName: string = '';
...
 const batcherUri: URI = ENV.BATCHER_URI;
 
-const chainUri: URI = ENV.CHAIN_URI;
-const chainExplorerUri: URI = ENV.CHAIN_EXPLORER_URI;
-const chainId: number = ENV.CHAIN_ID;
-const chainName: string = ENV.CHAIN_NAME;
-const chainCurrencyName: string = ENV.CHAIN_CURRENCY_NAME;
-const chainCurrencySymbol: string = ENV.CHAIN_CURRENCY_SYMBOL;
-const chainCurrencyDecimals: number = ENV.CHAIN_CURRENCY_DECIMALS;
+const [chainName, config] = await GlobalConfig.mainEvmConfig();
```
SebastienGllmt, line 20: FYI this will cause issues when called from frontends because `readFile` is not available from web browsers which is the same problem we had with `SECURITY_NAMESPACE` (now that I think of it, maybe `SECURITY_NAMESPACE` should be merged into the same file as this config)

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -82,10 +79,14 @@ async function runPresync(
...
-      doLog(`${JSON.stringify(presyncBlockHeight)}`);
+    for (const network of Object.keys(networks)) {
+      if (upper[network] > presyncBlockHeight[network]) {
+        doLog(
+          `${network}: ${JSON.stringify(presyncBlockHeight[network])}-${JSON.stringify(upper[network])}`
+        );
+      } else {
+        doLog(`${network}: ${JSON.stringify(presyncBlockHeight[network])}`);
+      }
     }
```
SebastienGllmt, line 90: What is this log even supposed to mean? If this is really something we want to show users on startup, we should make it clearer what this is supposed to be

packages/engine/paima-funnel/src/funnels/evm/funnel.ts
```diff
@@ -0,0 +1,557 @@
...
+        chainData.internalEvents = [];
+      }
+
+      // it's technically possible for this to be null, because there may not be
+      // a block of the sidechain in between a particular pair of blocks or the
+      // original chain.
+      //
+      // in this case it could be more optimal to set the block number here to
+      // the one in the next block, but it shouldn't make much of a difference.
+      // TODO: is that true?
```
SebastienGllmt, line 244: TODO?

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: Well, but all of these things currently have defaults... So I just kept that behavior. I took the defaults from the `ENV` class.

packages/paima-sdk/paima-mw-core/src/state.ts
ecioppettini: Right, I was concerned about that, but I wasn't sure what's included in browsers and what not. But what's the fix for that? Should I keep the ENV variables for this case? They could be filled to the main network settings after all.

packages/engine/paima-runtime/src/runtime-loops.ts
ecioppettini: The log was already there, it says what's the range that is going to be fetched in that presync round. I guess I could add a `fetching blocks in range: ` to the msg or something.

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: not really... But I have some weird issue with the build sometimes because of the `doLog` import, and the `ENV` import. But it doesn't happen all the time, and I couldn't find a way to fix it yet. For some reason it complains in paima-runtime about imports not being under the rootDir or something like that.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -257,18 +262,24 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
         basePromise,
       ]);
 
-      let grouped = groupCdeData(Network.CARDANO, arg.from, arg.to, carpEvents);
+      let grouped = groupCdeData(
+        this.chainName,
+        ConfigNetworkType.CARDANO,
```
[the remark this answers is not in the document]
ecioppettini: Well, the `carp` funnel can't really have another type. In reality I doubt there is a reason to have two cardano networks, but this part was mostly to be consistent. So I think this is indeed fine.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -422,7 +436,8 @@ async function readDataInternal(
   );
 
   let grouped = groupCdeData(
-    Network.EVM,
+    chainName,
+    ConfigNetworkType.CARDANO,
```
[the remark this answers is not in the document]
ecioppettini: This is a good question. I need to re-check actually.

I guess it was `EVM` because it's part of the sync, and not the presync. The thing is, it doesn't really change anything. This function returns `PresyncChainData`, and the network type field is used for the presync processing function, but when merging this with the `ChainData` from the evm chain during the sync it's discarded. The same thing happens with the `chainName` actually...

Maybe these fields should be optional? Or maybe `composeChainData` shouldn't actually take `PresyncChainData[]`, but a more specific type.

packages/engine/paima-funnel/src/funnels/block/funnel.ts
```diff
@@ -9,16 +9,27 @@ import type { FunnelSharedData } from '../BaseFunnel.js';
 import { RpcCacheEntry, RpcRequestState } from '../FunnelCache.js';
 import type { PoolClient } from 'pg';
 import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 12: This breaks the build

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -28,6 +29,7 @@ import { Routes } from '@dcspark/carp-client/shared/routes';
 import { FUNNEL_PRESYNC_FINISHED, InternalEventType } from '@paima/utils';
 import { CarpFunnelCacheEntry } from '../FunnelCache.js';
 import { getCardanoEpoch } from '@paima/db';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 32: this breaks as well

packages/engine/paima-funnel/src/funnels/evm/funnel.ts
```diff
@@ -0,0 +1,557 @@
...
+import type { FunnelSharedData } from '../BaseFunnel.js';
+import {
+  EvmFunnelCacheEntry,
+  EvmFunnelCacheEntryState,
+  RpcCacheEntry,
+  RpcRequestState,
+} from '../FunnelCache.js';
+import type { PoolClient } from 'pg';
+import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 26: this breaks the build

packages/engine/paima-funnel/src/index.ts
```diff
@@ -9,6 +15,9 @@ import { BlockFunnel } from './funnels/block/funnel.js';
 import type { FunnelSharedData } from './funnels/BaseFunnel.js';
 import { FunnelCacheManager } from './funnels/FunnelCache.js';
 import { wrapToCarpFunnel } from './funnels/carp/funnel.js';
+import { wrapToEvmFunnel } from './funnels/evm/funnel.js';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 19: this breaks the build

packages/engine/paima-funnel/src/utils.ts
```diff
@@ -1,8 +1,9 @@
 import type { ChainData, ChainDataExtensionDatum, PresyncChainData } from '@paima/sm';
-import type { Network } from '@paima/utils';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading';
```
SebastienGllmt, line 2: this breaks the build

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -15,6 +15,8 @@ import {
 import { cleanNoncesIfTime } from './nonce-gc.js';
 import type { PoolClient } from 'pg';
 import { FUNNEL_PRESYNC_FINISHED } from '@paima/utils';
+import type { CardanoConfig } from '@paima/utils/src/config/loading.js';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 19: this breaks the build

packages/engine/paima-sm/src/index.ts
```diff
@@ -44,6 +43,8 @@ import type {
   GameStateMachineInitializer,
   InternalEvent,
 } from './types.js';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 46: this breaks the build

packages/engine/paima-sm/src/types.ts
```diff
@@ -12,30 +12,38 @@ import type {
   PaimaERC721Contract,
   OldERC6551RegistryContract,
   ERC6551RegistryContract,
-  Network,
   InternalEventType,
 } from '@paima/utils';
 import { Type } from '@sinclair/typebox';
 import type { Static } from '@sinclair/typebox';
+import { ConfigNetworkType } from '@paima/utils/src/config/loading';
```
SebastienGllmt, line 19: this breaks the build

packages/paima-sdk/paima-utils/src/config/loading.ts
SebastienGllmt: Imports not being under the root dir is caused by the `import ... from .../src/...` issue I mentioned (since instead of importing a package, you're now importing a JS file that is under a totally different folder)

packages/engine/paima-runtime/src/runtime-loops.ts
SebastienGllmt: Yeah I know this was added in a previous commit, but every time I launch paima engine I always wonder what this log was meant to be, so best to be more descriptive

packages/paima-sdk/paima-mw-core/src/state.ts
SebastienGllmt: as we discussed in the call, it may not be required for most frontend if it fallsback to the .env file the basic stuff which is not great behavior, but not bad as a temporary solution. We just have to make sure that `mainEvmConfig` doesn't crash the app by trying to call `readFile` when running from the frontend and properly document this temporary behavior

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: I think we could forget about this unless someone asks for it, tbh. I personally made the mistake in the past of naming `extensions.yaml` instead of `extensions.yml`, since I think that's the most common one, but it doesn't really change much.

ecioppettini:
I'm going to post here a script that I've been using for testing this, just in case. Since otherwise I'm not sure where I would push this, but I don't want to lose it

I'm running it from the paima-game-templates repo/directory. It starts two anvil chains and pupeteers the block production timestamps in order to generate the cases that I wanted. It doesn't really have any assertions, I've been just looking at the logs.

**NOTE** I recommend manually backing up any settings before running, since it would delete them.

```sh
#!/usr/bin/env sh

set -e

trap 'kill $(jobs -p)' EXIT

anvil --port 8545 --chain-id 31337 --timestamp 0 >> /dev/null &
anvil --port 8546 --chain-id 31338 --timestamp 0 >> /dev/null &

forge init my-erc20 --no-commit --force
forge install openzeppelin/openzeppelin-contracts --no-commit

cd my-erc20

cat <<EOF > src/CustomErc20.sol
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

import "lib/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol";

contract CustomERC20 is ERC20 {
    constructor() ERC20("CustomErc20", "TOKEN") {
        _mint(msg.sender, 1000);
    }
} 
EOF

forge build

function lastBlock1() {
  TS=$(cast block 'latest' --rpc-url http://localhost:8545 | rg 'timestamp' | awk -F' +' '{ print $2 }')
  echo "Timestamp 1: $TS"
}

function lastBlock2() {
  TS=$(cast block 'latest' --rpc-url http://localhost:8546 | rg 'timestamp' | awk -F' +' '{ print $2 }')
  echo "Timestamp 2: $TS"
}

cast rpc evm_setNextBlockTimestamp 1 --rpc-url http://localhost:8545 >> /dev/null
cast rpc evm_setNextBlockTimestamp 1 --rpc-url http://localhost:8546 >> /dev/null

TOKEN1=$(forge create --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 src/CustomErc20.sol:CustomERC20 | grep "Deployed to" | cut -d':' -f2 | cut -d' ' -f2)
TOKEN2=$(forge create --rpc-url http://localhost:8546 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 src/CustomErc20.sol:CustomERC20 | grep "Deployed to" | cut -d':' -f2 | cut -d' ' -f2)

USER=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
TARGET=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC

cast rpc anvil_impersonateAccount $USER --rpc-url http://localhost:8545
cast rpc anvil_impersonateAccount $USER --rpc-url http://localhost:8546


######## START SIDE CHAIN ######## 

######## START PRESYNC EVENTS ######## 

# block 2
cast rpc evm_setNextBlockTimestamp 2 --rpc-url http://localhost:8546

cast send $TOKEN2 --unlocked --from $USER "transfer(address,uint256)" $TARGET 2 --rpc-url http://localhost:8546

# block 3
cast rpc evm_setNextBlockTimestamp 3 --rpc-url http://localhost:8546

cast send $TOKEN2 --unlocked --from $USER "transfer(address,uint256)" $TARGET 4 --rpc-url http://localhost:8546

######## END PRESYNC EVENTS ######## 

# block 4
cast rpc evm_setNextBlockTimestamp 4 --rpc-url http://localhost:8546

cast send $TOKEN2 --unlocked --from $USER "transfer(address,uint256)" $TARGET 5 --rpc-url http://localhost:8546

######## END SIDE CHAIN ######## 

######## START MAIN CHAIN ######## 

# block 2
cast rpc evm_setNextBlockTimestamp 3 --rpc-url http://localhost:8545

cast rpc evm_mine --rpc-url http://localhost:8545 >> /dev/null

# block 3
cast rpc evm_setNextBlockTimestamp 5 --rpc-url http://localhost:8545

cast send $TOKEN1 --unlocked --from $USER "transfer(address,uint256)" $TARGET 1 --rpc-url http://localhost:8545

######## END MAIN CHAIN ######## 

# block 5

# Mine a block so that we can know we have everything in the needed range
cast rpc evm_setNextBlockTimestamp 7 --rpc-url http://localhost:8546

cast rpc evm_mine --rpc-url http://localhost:8546 >> /dev/null

# DEPLOY PAIMA L2 CONTRACT

cd ..

cast rpc evm_setNextBlockTimestamp 6 --rpc-url http://localhost:8545
PAIMA_L2=$(forge create --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --hardhat --root contracts/paima-l2-contract src/contract/PaimaL2Contract.sol:PaimaL2Contract --constructor-args 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 0 | grep "Deployed to" | cut -d':' -f2 | cut -d' ' -f2)

echo "Paima L2 Contract deployed"

lastBlock1
lastBlock2

# set automining to verify that the sync is paused if the other chain stops
cast rpc evm_setIntervalMining 4 --rpc-url http://localhost:8545 >> /dev/null && echo "Set automining"

NETWORK=localhost

if test -f config.$NETWORK.yml; then
  if ! test -f config.$NETWORK.yml; then
    cp config.$NETWORK.yml config.$NETWORK.yml.bak
  fi
fi

cat <<EOF > config.$NETWORK.yml
Anvil1:
  type: evm-main
  chainUri: 'http://localhost:8545'
  chainId: 31337
  chainCurrencyName: 'ETH'
  chainCurrencySymbol: 'ETH'
  chainCurrencyDecimals: 18
  blockTime: 2
  paimaL2ContractAddress: '$PAIMA_L2'

Anvil2:
  type: evm-other
  chainUri: 'http://localhost:8546'
  chainId: 31338
  chainCurrencyName: 'ETH'
  chainCurrencySymbol: 'ETH'
  chainCurrencyDecimals: 18
  blockTime: 2
  presyncStepSize: 100
EOF

if test -f extensions.yml; then
  if ! test -f extensions.yml; then
    cp extensions.yml extensions.yml.bak
  fi
fi

cat <<EOF > extensions.yml
extensions:
  - name: "TOKEN1"
    type: "erc20"
    contractAddress: "$TOKEN1"
    startBlockHeight: 1
    network: "Anvil1"

  - name: "TOKEN2"
    type: "erc20"
    contractAddress: "$TOKEN2"
    startBlockHeight: 1
    network: "Anvil2"
EOF

trap 'kill $(jobs -p); rm -r my-erc20; rm extensions.yml; rm config.localhost.yml' EXIT

cd chess

npm run database:reset >> /dev/null
(npm run database:up >> /dev/null) &

cd ..

sleep 5s

sed -i 's/^START_BLOCKHEIGHT=.*/START_BLOCKHEIGHT=2/g' .env.localhost

echo "Starting Paima Engine"

./dev-paima-engine-linux run 
```

packages/engine/paima-funnel/src/funnels/evm/funnel.ts
ecioppettini: I think that was indeed true

packages/paima-sdk/paima-mw-core/src/state.ts
ecioppettini: I reverted this file to use the env file settings. I'm not sure if the change would also be needed somewhere else.

ecioppettini:
Similarly for cardano (just to check this didn't break the carp funnel). A bit messier since it mocks carp.

```sh
#!/usr/bin/env sh

set -e

trap 'kill $(jobs -p)' EXIT

BASE_TIMESTAMP=1666656000

timestamp() {
 echo $(($BASE_TIMESTAMP + 200 + $1));
}

anvil --port 8545 --chain-id 31337 --timestamp $(timestamp 1)  >> /dev/null &

echo "Network deployed"

sleep 2s

####### Mock carp to have some stuff

node <<EOF &
const http = require('http');

const routes = {
  '/asset/utxos': (req, res) => {
    if (req.method === 'POST') {
      let body = '';
      req.on('data', chunk => {
        body += chunk.toString();
      });
      req.on('end', () => {
        res.writeHead(200, { 'Content-Type': 'text/json' });

        const input = JSON.parse(body);

        const data = [{
          txId: '00000000000000000000000000000000',
          slot: 1,
          paymentCred: "00000000000000000000000000000000",
          assetName: "111111111111111111",
          policyId: "00000000000000000000000000000000",
          cip14Fingerprint: input.fingerprints[0],
          amount: "100",
          utxo: {
            index: 0,
            tx: "00000000000000000000000000000000"
          }
        },
        {
          txId: '00000000000000000000000000000000',
          slot: 5,
          paymentCred: "11111111111111111111111111111111",
          assetName: "111111111111111111",
          policyId: "00000000000000000000000000000000",
          cip14Fingerprint: input.fingerprints[0],
          amount: "120",
          utxo: {
            index: 0,
            tx: "11111111111111111111111111111111"
          }
        }
        ];

        res.end(JSON.stringify(data.filter(d => d.slot >= input.range.minSlot && d.slot < input.range.maxSlot), undefined, '\t'));
      });
    } else {
      res.writeHead(405, { 'Content-Type': 'text/plain' });
      res.end('Method Not Allowed');
    }
  },

  '/block/latest': (req, res) => {
    if (req.method === 'POST') {
      let body = '';
      req.on('data', chunk => {
        body += chunk.toString();
      });
      req.on('end', () => {
        const data = {
          block: {
            slot: 15,
            epoch: 0,
            height: 15,
            hash: '00000000000000000000000000000000',
            era: 0
          }
        };

        res.writeHead(200, { 'Content-Type': 'text/json' });
        res.end(JSON.stringify(data, undefined, '\t'));
      });
    } else {
      res.writeHead(405, { 'Content-Type': 'text/plain' });
      res.end('Method Not Allowed');
    }
  }
};

const server = http.createServer((req, res) => {
  const routeHandler = routes[req.url];
  if (routeHandler) {
    routeHandler(req, res);
  } else {
    res.writeHead(404, { 'Content-Type': 'text/plain' });
    res.end('Not Found');
  }
});

server.listen(3000, () => {});
EOF

sleep 1s

echo "Carp mock deployed"

####### DEPLOY PAIMA L2 CONTRACT

echo "Deploying Paima L2 Contract"

cast rpc evm_setNextBlockTimestamp $(timestamp 2) --rpc-url http://localhost:8545 >> /dev/null

PAIMA_L2=$(forge create --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --hardhat --root contracts/paima-l2-contract src/contract/PaimaL2Contract.sol:PaimaL2Contract --constructor-args 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 0 | grep "Deployed to" | cut -d':' -f2 | cut -d' ' -f2)

echo "Paima L2 Contract deployed"

cast rpc evm_setNextBlockTimestamp $(timestamp 3) --rpc-url http://localhost:8545 >> /dev/null

####### Deploy ERC20 

forge init my-erc20 --no-commit --force
forge install openzeppelin/openzeppelin-contracts --no-commit

pushd my-erc20

cat <<EOF > src/CustomErc20.sol
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

import "lib/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol";

contract CustomERC20 is ERC20 {
    constructor() ERC20("CustomErc20", "TOKEN") {
        _mint(msg.sender, 1000);
    }
} 
EOF

forge build

TOKEN1=$(forge create --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 src/CustomErc20.sol:CustomERC20 | grep "Deployed to" | cut -d':' -f2 | cut -d' ' -f2)

USER=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
TARGET=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC

popd

###### Impersonate accounts

cast rpc anvil_impersonateAccount $USER --rpc-url http://localhost:8545 >> "Impersonated account on network1"

######## START MAIN CHAIN ######## 

# block 3
cast rpc evm_setNextBlockTimestamp $(timestamp 4) --rpc-url http://localhost:8545 >> /dev/null

cast rpc evm_mine --rpc-url http://localhost:8545 >> /dev/null

# block 4
cast rpc evm_setNextBlockTimestamp $(timestamp 6) --rpc-url http://localhost:8545 >> /dev/null

cast send $TOKEN1 --unlocked --from $USER "transfer(address,uint256)" $TARGET 1 --rpc-url http://localhost:8545 >> /dev/null

# block 5
cast rpc evm_setNextBlockTimestamp $(timestamp 7) --rpc-url http://localhost:8545 >> /dev/null

cast send $TOKEN1 --unlocked --from $USER "transfer(address,uint256)" $TARGET 1 --rpc-url http://localhost:8545 >> /dev/null

# block 6
cast rpc evm_setNextBlockTimestamp $(timestamp 9) --rpc-url http://localhost:8545 >> /dev/null
cast rpc evm_mine --rpc-url http://localhost:8545 >> /dev/null

# block 7

cast rpc evm_setNextBlockTimestamp $(timestamp 10) --rpc-url http://localhost:8545 >> /dev/null
cast send $TOKEN1 --unlocked --from $USER "transfer(address,uint256)" $TARGET 1 --rpc-url http://localhost:8545 >> /dev/null

# block 8
# cast rpc evm_setNextBlockTimestamp $(timestamp 300) --rpc-url http://localhost:8545 >> /dev/null
# cast send $TOKEN1 --unlocked --from $USER "transfer(address,uint256)" $TARGET 1 --rpc-url http://localhost:8545 >> /dev/null

######## END MAIN CHAIN ######## 

NETWORK=localhost

if test -f config.$NETWORK.yml; then
  if ! test -f config.$NETWORK.yml; then
    cp config.$NETWORK.yml config.$NETWORK.yml.bak
  fi
fi

cat <<EOF > config.$NETWORK.yml
Anvil1:
  type: evm-main
  chainUri: 'http://localhost:8545'
  chainId: 31337
  chainCurrencyName: 'ETH'
  chainCurrencySymbol: 'ETH'
  chainCurrencyDecimals: 18
  blockTime: 2
  paimaL2ContractAddress: '$PAIMA_L2'


Cardano:
  type: cardano
  carpUrl: http://localhost:3000
  network: preview
  confirmationDepth: 10
EOF

if test -f extensions.yml; then
  if ! test -f extensions.yml; then
    cp extensions.yml extensions.yml.bak
  fi
fi


cat <<EOF > extensions.yml
extensions:
  - name: "TOKEN1"
    type: "erc20"
    contractAddress: "$TOKEN1"
    startBlockHeight: 1
    network: "Anvil1"

  - name: "CARDANO-ASSET1"
    type: cardano-delayed-asset
    fingerprints:
      - "asset1c43p68zwjezc7aaaaaaaaaaaaaaaaaaaaaaaaa"
    startSlot: 0
    network: Cardano
EOF

trap 'kill $(jobs -p); rm -r my-erc20; rm extensions.yml; rm config.localhost.yml; pushd chess/db/docker; docker compose down; popd; sleep 5s' EXIT

pushd chess

npm run database:reset >> /dev/null

pushd db/docker

(docker compose up >> /dev/null) &

popd
popd

sleep 5s

sed -i 's/^START_BLOCKHEIGHT=.*/START_BLOCKHEIGHT=3/g' .env.$NETWORK

echo "Starting Paima Engine"

TMPFILE=$(mktemp)

./dev-paima-engine-linux run >> $TMPFILE & PAIMA_PID=$!; sleep 5s; kill -INT %+; wait $PAIMA_PID

echo "Logs"

cat $TMPFILE

export PGDATABASE="postgres"
export PGUSER="postgres"
export PGPASSWORD="postgres"
export PGHOST="localhost"
export PGPORT=5432

psql -c 'SELECT * FROM cde_erc20_data;'
psql -c 'SELECT * FROM cde_cardano_asset_utxos;'

###### Assertions

assertEq() {
  if [ $1 -ne $2 ]; then
      echo "Assertion failed: $3. Expected $1 found $2."
      exit 1
  fi
}

ANVIL1_BALANCE=$(psql -c "SELECT * FROM cde_erc20_data WHERE cde_id=0 AND wallet_address='0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc';" | tail -3 | head -1 | tr -d ' '| awk -F '|' '{ print $3 }')

CARDANO1_BALANCE=$(psql -c "SELECT amount FROM cde_cardano_asset_utxos WHERE cde_id=1 AND address='00000000000000000000000000000000';" | tail -3 | head -1 | tr -d ' '| awk -F '|' '{ print $1 }')
CARDANO2_BALANCE=$(psql -c "SELECT amount FROM cde_cardano_asset_utxos WHERE cde_id=1 AND address='11111111111111111111111111111111';" | tail -3 | head -1 | tr -d ' '| awk -F '|' '{ print $1 }')

assertEq $ANVIL1_BALANCE 3 "Wrong balance on main chain"

assertEq $CARDANO1_BALANCE 100 "Wrong balance on cardano address 1"
assertEq $CARDANO2_BALANCE 120 "Wrong balance on cardano address 2"

## asserts over logs

assertEq $(grep '2 CDE events in block #4' $TMPFILE >> /dev/null; echo $?) 0 "Wrong grouping for block 2"
assertEq $(grep 'Processed 1 CDE events in slot #1' $TMPFILE >> /dev/null; echo $?) 0 "Error in presync"


echo ''
echo -e "                            \e[1;32mSuccess\e[0m"
echo ''
```

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -167,7 +188,8 @@ async function runPresyncRound(
     await gameStateMachine.presyncProcess(dbTx, presyncData);
   }
 
-  const cardanoFrom = from.find(arg => arg.network === Network.CARDANO);
+  // TODO: use the non-hardcoded name
```
SebastienGllmt, line 191: todo?

packages/engine/paima-funnel/src/funnels/emulated/funnel.ts
```diff
@@ -249,6 +251,8 @@ export class EmulatedBlocksFunnel extends BaseFunnel {
       return newEntry.processingQueue;
     })();
 
+    const [chainName, _config] = await GlobalConfig.mainEvmConfig();
```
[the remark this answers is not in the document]
ecioppettini: Well, there is always a main evm funnel at the base of the stack of funnels. And the actual "blocks" that are received by this funnel are essentially the ones for the main chain. That said, I'm actually not sure how this is supposed to work, especially during the presync, since I haven't really used the emulated funnel that much.

In this case I'm not really using the config, I'm just using the chain name, and this is only used in order to know which of the events that come from the earlier funnels needs to be used.

That said, maybe this is incorrect? Maybe there should be an emulated funnel PER evm funnel instead? And each one should process only the blocks from that particular network. Is that what you are suggesting?

packages/engine/paima-funnel/src/funnels/evm/funnel.ts
```diff
@@ -0,0 +1,565 @@
...
+  RpcRequestState,
+} from '../FunnelCache.js';
+import type { PoolClient } from 'pg';
+import { FUNNEL_PRESYNC_FINISHED, ConfigNetworkType } from '@paima/utils';
+import { getMultipleBlockData } from '../../reading.js';
+import { getLatestProcessedCdeBlockheight } from '@paima/db';
+
+const GET_BLOCK_NUMBER_TIMEOUT = 5000;
+
+export class EvmFunnel extends BaseFunnel implements ChainFunnel {
```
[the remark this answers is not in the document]
ecioppettini: what would work though? `parallelEvmFunnel` maybe? `companionEvmChainFunnel` ? 

I didn't want to name it `sidechainEvmFunnel` since I feel that may be confusing because the word sidechain already means something.

`PairedEvmFunnel` maybe?

SebastienGllmt (approved):

Issue #262, opened 2023-12-05
Cardano transfer CDE


SebastienGllmt:
**Expected use-case**: detect if somebody sent ADA to us to buy something in the game. Use tx metadata / datum to specify what they are buying

This CDE takes in a payment credential (could be a payment key hash, could be a script hash) and monitors all ADA sent to that credential.

The main question is: what data does Paima get to see when it happens?
1. It can't just be the tx cbor, because then you don't know the context of the inputs (other than tx hash and id)
2. It can't be the Plutus context since it doesn't include tx metadata (which we want)

So we have two choices:
1. Give the entire tx context + extra information about the tx inputs (in Carp these are two separate queries, but we'd have to modify the tx history endpoint to give both at once)
2. Give some subset of the tx info that we deem sufficient

Picking (2) maybe is useful for most basic case possible for people who don't want to deal with the complexity of (1). In which case, I would give the following:

Update to the state machine should see the full transaction CBOR from Cardano. The reason I think we have to do this (instead of providing a simpler interface) is because a lot of use-cases may want 
1. List of outputs the funds were sent to that contain the stake credential
2. List of all inputs
3. Transaction metadata
4. Output datum for the UTXO created (hash or inline)

One thing mentioned by Eugene is that we can actually remove "list of inputs" from the requirements if we have users specify the address to credit the data to in the game as part of the metadata (which we probably want for a lot of cases anyway)

**Important note**: there may be multiple outputs with the same address in the transaction, so this CDE should not run "per output", but rather once for the transaction in general

SebastienGllmt:
See #280

PR #308, opened 2024-03-06
cardano mint burn

ecioppettini:
this is branched from https://github.com/PaimaStudios/paima-engine/pull/307 because I implemented this on top of the paginated endpoint, since it was a bit easier to do it that way.

it wouldn't be too much effort to rebase this on top of current main, if we need it for some reason, but it doesn't seem to be worth it.

Current config looks like this:

```yml
extensions:
  - name: "CARDANO-MINT-BURN"
    type: cardano-mint-burn
    policyIds: 
      - "0fc891fb7368d3b7c7b88815c203fda0d6862b0f1d797222672e91fe"
    startSlot: 722300
    stopSlot: 733216
    scheduledPrefix: cmb
    network: Cardano
```

Inputs to the state transition look like:

```
cmb|5a247451a8e02ea80f75893a61ed70a1ae65e0cd552f45fe993c3621efa16011|a1636d73678f78264175746f2d4c6f6f702d5472616e73616374696f6e202335353034333920627920415441444160783a4c6976652045706f636820392c207765206861766520303031682030346d20313973206c65667420756e74696c20746865206e657874206f6e6578384974277320446f6e6e657273746167202d203033204e6f76656d6265722032303232202d2032333a35353a343120696e2041757374726961607837f09f9388205468652063757272656e7420414441205072696365206f6e204b72616b656e2069733a2024302e3339303635202f20414441607820412072616e646f6d205a656e2d51756f746520666f7220796f753a20f09f998f7838466f6f6c206d65206f6e63652c207368616d65206f6e20796f753b20666f6f6c206d652074776963652c207368616d65206f6e206d652e2072202d204368696e6573652050726f766572626078374e6f64652d5265766973696f6e3a203935306334653232323038366665643563613533353634653634323433346365393330376230623960782d507265566965772d436861696e20697320617765736f6d652c206861766520736f6d652066756e2120f09f988d7819204265737420726567617264732c204d617274696e203a2d29|{"0fc891fb7368d3b7c7b88815c203fda0d6862b0f1d797222672e91fe":{"4d79416d617a696e674e4654":"1"}}
```

And I'm currently using this for parsing:

```ts
const cardanoMint: ParserRecord<CardanoMint> = {
  txId: PaimaParser.NCharsParser(0, 64),
  metadata: PaimaParser.OptionalParser(null, PaimaParser.RegexParser(/[a-f0-9]*/)),
  assets: (keyName: string, input: string) => {
    return JSON.parse(input);
  },
};
```

```ts
export interface CardanoMint {
  txId: string;
  metadata: string | null;
  assets: { asset: { policyId: string; assetName: string }; amount: string };
}
```

Which is similar to the stuff in #262

### DIFF
--- a/packages/engine/paima-sm/src/cde-cardano-mint-burn.ts
+++ b/packages/engine/paima-sm/src/cde-cardano-mint-burn.ts
@@ -1,0 +1,31 @@
+import { ENV } from '@paima/utils';
+import { cdeCardanoMintBurnInsert, createScheduledData } from '@paima/db';
+import type { SQLUpdate } from '@paima/db';
+import type { CdeCardanoMintBurnDatum } from './types.js';
+
+export default async function processDatum(
+  cdeDatum: CdeCardanoMintBurnDatum
+): Promise<SQLUpdate[]> {
+  const cdeId = cdeDatum.cdeId;
+  const prefix = cdeDatum.scheduledPrefix;
+  const txId = cdeDatum.payload.txId;
+  const assets = JSON.stringify(cdeDatum.payload.assets);
+  const metadata = cdeDatum.payload.metadata || undefined;
+
+  const scheduledBlockHeight = Math.max(cdeDatum.blockNumber, ENV.SM_START_BLOCKHEIGHT + 1);
+  const scheduledInputData = `${prefix}|${txId}|${metadata}|${assets}`;
+
+  const updateList: SQLUpdate[] = [
+    createScheduledData(scheduledInputData, scheduledBlockHeight),
+    [
+      cdeCardanoMintBurnInsert,
+      {
+        cde_id: cdeId,
+        tx_id: txId,
+        metadata: metadata,
+        assets: cdeDatum.payload.assets,
+      },
+    ],
+  ];
+  return updateList;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -26295,7 +26295,7 @@
       "name": "@paima/batcher-address-validator",
       "version": "0.1.0",
       "dependencies": {
-        "@dcspark/carp-client": "^3.0.0",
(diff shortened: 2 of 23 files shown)


SebastienGllmt:
As a reminder, we'll want to add this to the docs as well

packages/engine/paima-sm/src/cde-cardano-mint-burn.ts
```diff
@@ -0,0 +1,31 @@
...
+  cdeDatum: CdeCardanoMintBurnDatum
+): Promise<SQLUpdate[]> {
+  const cdeId = cdeDatum.cdeId;
+  const prefix = cdeDatum.scheduledPrefix;
+  const txId = cdeDatum.payload.txId;
+  const assets = JSON.stringify(cdeDatum.payload.assets);
+  const metadata = cdeDatum.payload.metadata || undefined;
+
+  const scheduledBlockHeight = Math.max(cdeDatum.blockNumber, ENV.SM_START_BLOCKHEIGHT + 1);
+  const scheduledInputData = `${prefix}|${txId}|${metadata}|${assets}`;
```
SebastienGllmt, line 16: what happens if `metadata` contains a `|` in it? Is that possible in the encoding we use?

ecioppettini: it's not really possible because right now it's just hex (of the cbor format). If we decode it to a json I guess it should be possible. Do we want to do that?

SebastienGllmt: I think hex is fine 👍

SebastienGllmt (approved):
I need to review the docs still, but this PR itself makes sense

PR #309, opened 2024-03-08
Updated inverse projection

SebastienGllmt:
This updates the contracts for inverse projection based on https://github.com/PaimaStudios/PRC/pull/4

### DIFF
--- a/packages/contracts/evm-contracts/contracts/token/IInverseAppProjectedNft.sol
+++ b/packages/contracts/evm-contracts/contracts/token/IInverseAppProjectedNft.sol
@@ -3,7 +3,8 @@ pragma solidity ^0.8.13;
 
 import {IInverseProjectedNft} from "./IInverseProjectedNft.sol";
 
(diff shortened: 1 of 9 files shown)


packages/contracts/evm-contracts/contracts/token/IInverseAppProjectedNft.sol
```diff
@@ -3,14 +3,14 @@ pragma solidity ^0.8.13;
 
 import {IInverseProjectedNft} from "./IInverseProjectedNft.sol";
 
-/// @dev A standard ERC721 that accepts calldata in the mint function for any initialization data needed in a Paima dApp.
+/// @dev A standard ERC721 that can be freely minted and stores an unique <minter, userTokenId> pair (used in tokenURI) when minted.
```
SebastienGllmt, line 6: Do we have to update these comments in PRC3 as well?

matejos: Yes good point, I've updated it here and also there.

PR #311, opened 2024-03-10
parallel evm funnel: after presync start from startBlockHeight

ecioppettini:
There is currently some code that saves the latest height in order to know from where to start syncing on restart on the parallel chain. The issue is that during the presync this only gets updated if there is an event in the range. So potentially the presync may end up in a block far way from the `startingBlock`, and the sync starts from that point. Even if we already know there is nothing useful there. This sets things in a way that we never start the sync from a point that falls in the presync range.

Also while figuring what was happening I noticed that the funnel is updating the latest chain block much more than necessary, and it's a bit slow, so this optimizes that a bit.

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
@@ -131,15 +131,17 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
         cachedState.lastBlock = blocks[blocks.length - 1].blockNumber;
       }
 
-      // If we didn't break because there are no blocks in the chain then we
-      // need to wait for blocks to be produced.  But otherwise there is not
-      // much point in updating the value, since we may be really far away from
-      // the tip in the sync itself, and this adds a whole round trip.
-      if (to === latestBlock) {
-        while ((await this.updateLatestBlock()) === latestBlock) {
-          // wait for blocks to be produced
-          await delay(500);
-        }
+      // We reach this part of the code if after we fetch blocks we still aren't done syncing.
+      // There are 2 cases for this:
+      // 1. We're still far behind the tip and we're catching up
+      // For case (1), we can fetch more blocks right away (no need to update our cached latest block or delay)
+      if (to !== latestBlock) continue;
+      // 2. We're at the tip and we're waiting for more parallel chains blocks to be made to finalize the block
+      // For case (2), we try update our cached latest block number in case it's stale,
+      // and delay if we are truly waiting for more blocks to be made
+      while ((await this.updateLatestBlock()) === latestBlock) {
+        // wait for blocks to be produced
+        await delay(500);
       }
 
       // potentially we didn't fetch enough blocks in a single request in that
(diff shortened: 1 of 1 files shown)


packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
```diff
@@ -122,9 +126,15 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
         cachedState.lastBlock = blocks[blocks.length - 1].blockNumber;
       }
 
-      while ((await this.updateLatestBlock()) === latestBlock) {
-        // wait for blocks to be produced
-        await delay(500);
+      // If we didn't break because there are no blocks in the chain then we
+      // need to wait for blocks to be produced.  But otherwise there is not
+      // much point in updating the value, since we may be really far away from
+      // the tip in the sync itself, and this adds a whole round trip.
```
SebastienGllmt, line 132:
```suggestion
      // We reach this part of the code if after we fetch blocks we still aren't done syncing.
      // There are 2 cases for this:
      // 1. We're still far behind the tip and we're catching up
      // 2. We're at the tip and we're waiting for more parallel chains blocks to be made to finalize the block
      // For case (1), there is no need to delay
```

packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
```diff
@@ -122,9 +126,15 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
...
+      if (to === latestBlock) {
+        while ((await this.updateLatestBlock()) === latestBlock) {
+          // wait for blocks to be produced
+          await delay(500);
+        }
       }
```
SebastienGllmt, line 138:
```suggestion
      // We reach this part of the code if after we fetch blocks we still aren't done syncing.
      // There are 2 cases for this:
      // 1. We're still far behind the tip and we're catching up
      // For case (1), there is no need to do a network request or delay
      if (to !== latestBlock) continue;
      // 2. We're at the tip and we're waiting for more parallel chains blocks to be made to finalize the block
      // For case (2), we try update our cached latest block number in case it's stale,
      // and delay if we are truly waiting for more blocks to be made
      while ((await this.updateLatestBlock()) === latestBlock) {
        // wait for blocks to be produced
        await delay(500);
      }
```

SebastienGllmt: If you accept this suggestion, you also need to handle the 2 comments that come right after my suggested change (Github wouldn't let me modify them)

packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
```diff
@@ -85,7 +85,10 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
       );
 
       if (queryResults[0]) {
-        cachedState.lastBlock = queryResults[0].block_height;
+        cachedState.lastBlock = Math.max(
```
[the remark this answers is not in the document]
ecioppettini: It's fine because this is done inside `readData`, and that function is not called if the presync is not finished. And this cache entry is not used to decide that.

SebastienGllmt (approved):

PR #312, opened 2024-03-10
Parallel funnel edge-case fix

SebastienGllmt:
I believe I've found an edge-case that wasn't handled properly in the parallel funnel when multiple blocks have the same timestamp during the binary search. I created a fix for this (but definitely double-check my logic)

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
@@ -1,26 +1,13 @@
-import {
-  ENV,
-  EvmConfig,
-  doLog,
-  initWeb3,
-  logError,
-  timeout,
-  Web3,
-  delay,
-  InternalEventType,
-} from '@paima/utils';
+import type { EvmConfig, Web3 } from '@paima/utils';
+import { doLog, initWeb3, logError, timeout, delay, InternalEventType } from '@paima/utils';
 import type { ChainFunnel, ReadPresyncDataFrom } from '@paima/runtime';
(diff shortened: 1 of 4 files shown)


packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
```diff
@@ -542,7 +541,9 @@ async function findBlockByTimestamp(
 
     requests++;
 
-    if (Number(block.timestamp) < timestamp) {
+    // recall: there may be many blocks with the same targetTimestamp
+    // in this case, <= means we slowly increase `low` to return the most recent block with that timestamp
+    if (Number(block.timestamp) <= targetTimestamp) {
```
SebastienGllmt, line 546: This is the line to check (everything else in this PR is just cosmetic)

ecioppettini: Good point about the repeated timestamps, I forgot about that.

First of all, this changes the current behavior, but I think it may be more correct this way?

In practice I think which block we pick doesn't really matter that much, since the only thing that matters about the starting block that we find is that it is close enough, and that the search is deterministic.

Now, for the part about consecutive blocks with the same timestamp. I think this makes sense, although I found this comment a bit confusing because of the word `return`. 

> in this case, <= means we slowly increase `low` to return the most recent block with that timestamp

This is because I thought this was about return value of the function, but the function will actually return the _oldest_ block, if the next timestamp is also repeated. This is most likely nitpicking though.

ecioppettini (approved):

PR #313, opened 2024-03-11
fix parallel funnel block mapping issue

ecioppettini:
There are two issues here:

1. The cdeDatum blockNumber is not getting mapped to the main evm chain when using `internalReadDataSingle` (but it is when using `internalReadDataMultiple`). This can potentially lead to non scheduling inputs at the right time.
2. Currently the objects that we are using to map blocks from the parallel chain to the main chain are constructed _only_ with the timestamps obtained in the current call to `readData`. This doesn't correctly handle all the cases, since potentially the funnels can be configured to get 1 block from the wrapped funnel and a 100 of the other one. The fix here is basically to build these objects from the cached data, which includes data from the previous round, but was not needed.

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
+++ b/packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
@@ -220,7 +220,11 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
 
     if (toBlock === fromBlock) {
       doLog(`EVM CDE funnel ${this.config.chainId}: #${toBlock}`);
-      data = await this.internalReadDataSingle(fromBlock, chainData[0]);
+      data = await this.internalReadDataSingle(
+        fromBlock,
+        chainData[0],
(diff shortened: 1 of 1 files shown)


packages/engine/paima-funnel/src/funnels/parallelEvm/funnel.ts
```diff
@@ -198,10 +170,68 @@ export class ParallelEvmFunnel extends BaseFunnel implements ChainFunnel {
...
+    //    This chain pull: [1, 2, 3, 4]
+    // Round 2:
+    //    Base chain pull: [4]
+    //    This chain pull: [5]
+    //
+    //  We would want to merge this new base block with the last from the
+    // previous round (timestamp 4).
+    //
+    //  We know that this block it's still in the cached state, because we only
+    // removed stuff <= 3 (maxTimestamp). If we used the blocks from the current
```
SebastienGllmt, line 192:
```suggestion
    //  We know that this block is still in the cached state, because we only
    // removed blocks with timestamp <= 3 (maxTimestamp). If we used the blocks from the current
```

SebastienGllmt (approved):

PR #314, opened 2024-03-11
parallel evm funnel: fix internal events not propagated and incorret network in cde_tracking

ecioppettini:
Fixes some issues with the parallel funnel behavior.

1. This moves the network field currently in `ChainData` to the payload of each particular extension. This is because otherwise the CDE's in a parallel chain get indexed under the main chain, incorrectly, which later causes issue with the presync.
2. The emulated funnel doesn't currently propagate the internal events. This fixes that.
3. Also avoid storing scheduled data during the presync. This is a bit of an issue with the parallel funnel, because the block heights may be higher than the starting block, and get scheduled incorrectly. But also, it's not useful to have these things in the scheduled data.


SebastienGllmt (approved):

SebastienGllmt (approved):
accepting modulo our discussion on Slack which we need to decide and document

PR #315, opened 2024-03-12
evm-contracts: Use forge-std instead of excerpted files

matejos:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "packages/contracts/evm-contracts/lib/forge-std"]
+	path = packages/contracts/evm-contracts/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
```
SebastienGllmt, line 3: Is there no other way to do this? Submodules are always an absolute nightmare to work with

matejos: This is the default way forge is installed, but... I guess we could replace the submodule with the files directly

SebastienGllmt: Is there any downside to the approach we had before where things are inline? I would prefer that over having to update every build tool going forward to support submodules

matejos: I've ran into a few missing functions in the inline files, and I wasn't able to get them working just by adding them there. That's why I wanted to refactor into properly using forge-std. But in the end it is not a must-have, and if this would produce more problems than benefits, I'm fine with sunsetting this PR.

acedward:
We moved to Effectstream with evm update

PR #318, opened 2024-03-13
Orderbook DEX contract

matejos:
per https://github.com/PaimaStudios/tarochi-game/issues/67

### DIFF
--- a/packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
+++ b/packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
@@ -15,21 +15,6 @@ contract OrderbookDex is IOrderbookDex, ERC165, ReentrancyGuard {
     mapping(uint256 => Order) public orders;
     uint256 public ordersIndex;
 
-    event OrderCreated(
-        uint256 indexed orderId,
-        address indexed seller,
-        uint256 assetAmount,
-        uint256 price
-    );
-    event OrderFilled(
-        uint256 indexed orderId,
-        address indexed seller,
-        address indexed buyer,
-        uint256 assetAmount,
-        uint256 price
-    );
-    event OrderCancelled(uint256 indexed orderId);
-
     error OrderIsInactive(uint256 orderId);
     error Unauthorized();
 
--- a/packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
+++ b/packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
@@ -1,0 +1,102 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import {IOrderbookDex} from "./IOrderbookDex.sol";
+
+/// @notice Facilitates trading an asset that is living on a different app-chain.
+contract OrderbookDex is IOrderbookDex, ERC165, ReentrancyGuard {
+    using Address for address payable;
+
+    mapping(uint256 => Order) public orders;
+    uint256 public ordersIndex;
+
+    event OrderCreated(
+        uint256 indexed orderId,
+        address indexed seller,
+        uint256 assetAmount,
+        uint256 price
+    );
+    event OrderFilled(
+        uint256 indexed orderId,
+        address indexed seller,
+        address indexed buyer,
+        uint256 assetAmount,
+        uint256 price
+    );
+    event OrderCancelled(uint256 indexed orderId);
+
+    error OrderIsInactive(uint256 orderId);
+    error Unauthorized();
+
+    /// @notice Returns the current index of orders (index that a new sell order will be mapped to).
+    function getOrdersIndex() public view returns (uint256) {
+        return ordersIndex;
+    }
+
+    /// @notice Returns the Order struct information about order of specified `orderId`.
+    function getOrder(uint256 orderId) public view returns (Order memory) {
+        return orders[orderId];
+    }
+
+    /// @notice Creates a sell order for the specified `assetAmount` at specified `price`.
+    /// @dev The order is saved in a mapping from incremental ID to Order struct.
+    function createSellOrder(uint256 assetAmount, uint256 price) public {
+        Order memory newOrder = Order({
+            seller: payable(msg.sender),
+            assetAmount: assetAmount,
+            price: price,
+            active: true
+        });
+        orders[ordersIndex] = newOrder;
+        emit OrderCreated(ordersIndex, msg.sender, assetAmount, price);
+        ordersIndex++;
+    }
+
+    /// @notice Fills an array of orders specified by `orderIds`.
+    /// @dev Reverts if msg.value is less than the sum of orders' prices.
+    /// If msg.value is more than the sum of orders' prices, it should refund the difference back to msg.sender.
+    function fillSellOrders(uint256[] memory orderIds) public payable nonReentrant {
+        uint256 length = orderIds.length;
+        uint256 totalPaid;
+        for (uint256 i = 0; i < length; ) {
+            uint256 orderId = orderIds[i];
+            Order memory order = orders[orderId];
+            if (!order.active) {
+                revert OrderIsInactive(orderId);
+            }
+            order.seller.sendValue(order.price);
+            totalPaid += order.price;
+            orders[orderId].active = false;
+            emit OrderFilled(orderId, order.seller, msg.sender, order.assetAmount, order.price);
+            unchecked {
+                i++;
+            }
+        }
+        if (msg.value > totalPaid) {
+            payable(msg.sender).sendValue(msg.value - totalPaid);
+        }
+    }
+
+    /// @notice Cancels the sell order specified by `orderId`, making it unfillable.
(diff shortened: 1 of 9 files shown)


packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -0,0 +1,87 @@
...
+        ordersIndex++;
+    }
+
+    /// @notice Fills an array of orders specified by `orderIds`.
+    /// @dev Reverts if msg.value is less than the sum of orders' prices.
+    /// If msg.value is more than the sum of orders' prices, it should refund the difference back to msg.sender.
+    function fillSellOrders(uint256[] memory orderIds) public payable nonReentrant {
+        uint256 length = orderIds.length;
+        uint256 totalPaid;
+        for (uint256 i = 0; i < length; ) {
```
SebastienGllmt, line 51: Do you know the maximum number of orders we can fill at once? Probably we'll need to calculate this for the frontend

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -0,0 +1,87 @@
...
+            uint256 orderId = orderIds[i];
+            Order memory order = orders[orderId];
+            if (!order.active) {
+                revert OrderIsInactive(orderId);
+            }
+            order.seller.sendValue(order.price);
+            totalPaid += order.price;
+            orders[orderId].active = false;
+            emit OrderFilled(orderId, order.seller, msg.sender, order.assetAmount, order.price);
+            unchecked {
```
SebastienGllmt, line 61: this function needs to also create a new order if an order was partially filled.

Ex: if an order is to sell 5000 TGOLD
And you buy 3000 TGOLD

It should mark the 5000 order as inactive (to avoid concurrency issues) and create a new order with 2000 tgold

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -0,0 +1,87 @@
...
+    function createSellOrder(uint256 assetAmount, uint256 price) public {
+        Order memory newOrder = Order({
+            seller: payable(msg.sender),
+            assetAmount: assetAmount,
+            price: price,
+            active: true
+        });
+        orders[ordersIndex] = newOrder;
+        emit OrderCreated(ordersIndex, msg.sender, assetAmount, price);
+        ordersIndex++;
```
SebastienGllmt, line 42: Although an increasing index works, it will make the UI slow to update (recall: we had the same problem in the inverse projection where if the inverse projection is base-initiated it takes a while for the game node to update)

In this case however, we can't easily rely on app-side initiation solely because new UTXO entries are created when an order is partially filled (which has a new index)

To solve it though, I think we can take a variation of what we did for the inverse projection system:

1. Initiate your sell order on Tarochi (by placing in your gold to trade). This will give you a wallet-specific counter (instead of a global counter like we have here)
2. The RPC endpoint we query to know is `https://${rpcBase}/inverseProjection/${purpose}/${chainIdentifier}/` just like in PRC3, but with `purpose` set to `orderbook`
3. Unlike the NFT case where we queried for the file `${address}/${userTokenId}`, we instead query for `${address}/${orderId}/${amount}`. The endpoint then returns something like `{ aboveInitial: true }` if the amount provided is above the initial tgold deposited by the user in (1)

Then, like PRC3, we can create a function in this contract (one with the default RPC, one with the user RPC) that constructs the RPC URL from the Solidity contract.

With this configuration, we should end up where the Solidity contract provides an endpoint that always works right away to know if the order is real (has enough tgold). You simply check
1. if `active === true` in the Solidity contract
2. if `aboveInitial === true` from the RPC call

SebastienGllmt: Note that this requires essentially 2 IDs that we need to work with:
1. An ID for the initial orderID
2. A sub-ID for how far we are in a chain of partial orders

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
matejos: I didn't know we wanted to support partial fills. In that case:
1. Shouldn't we make it so that the seller can choose if they want their sell order to be single fill or if they want to allow partial fills? (afaik this is how it's done on whales.market for example)
2. In the case of partial fill, why would we have to discard the old order and create a new order? I understand that you're thinking in terms of UTxO model, but is this really necessary? Example: There is sell order for 1000 TGOLD. I submit a buy for 800 TGOLD but somebody else submits a buy of 600 TGOLD with higher gas fee, so by the time my transaction is being processed, 600 was already bought and so the sell order would only have 400 TGOLD left. I would I would much rather have my buy order fill for the 400 TGOLD (because it was a good price per token) than it to revert and me not getting anything and losing money on failed transaction gas fee. To be extra inclusive, we can make it an option for the buyer - "allow partial buy fill" (because maybe somebody really does want their buy order to either fill entirely or fail).

SebastienGllmt: I moved my comment on this topic here: https://github.com/PaimaStudios/PRC/pull/5#discussion_r1525212545

SebastienGllmt: > Shouldn't we make it so that the seller can choose if they want their sell order to be single fill or if they want to allow partial fills? (afaik this is how it's done on whales.market for example)

I don't think this is necessary. The reason whales.market does this is because users have to manually choose which order they want to fill, but we don't have that issue since it's all automated by the UI. At most we could do something to avoid dust (ex: an entry having so little tgold in it that it's not worth the tx fee to include it), but that's a little complicated

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -3,41 +3,59 @@
...
 /// @notice Facilitates trading an asset that is living on a different app-chain.
-contract OrderbookDex is IOrderbookDex, ERC165, ReentrancyGuard {
+contract OrderbookDex is IOrderbookDex, ERC1155Holder, ReentrancyGuard {
     using Address for address payable;
 
-    mapping(address => mapping(uint256 => Order)) public orders;
-    mapping(address => uint256) public sellersOrderId;
+    IInverseProjected1155 internal asset;
+    mapping(address => mapping(uint256 => Order)) internal orders;
+    mapping(address => uint256) internal sellersOrderId;
```
SebastienGllmt: we don't need any per-address counter in the OrderbookDex contract anymore because the ERC1155 contract does this for us.

To lookup whether or not a DEX order is real, you would do:
1. From the DEX contract, find the tokenID
2. Lookup the URI for that tokenID in the erc1155 inverse projection contract

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -3,41 +3,59 @@
...
-    function getOrder(address seller, uint256 orderId) external view returns (Order memory) {
+    function getOrder(address seller, uint256 orderId) public view virtual returns (Order memory) {
         return orders[seller][orderId];
     }
 
     /// @notice Creates a sell order with incremental seller-specific `orderId` for the specified `assetAmount` at specified `pricePerAsset`.
     /// @dev The order information is saved in a nested mapping `seller address -> orderId -> Order`.
     /// MUST emit `OrderCreated` event.
-    function createSellOrder(uint256 assetAmount, uint256 pricePerAsset) public {
+    function createSellOrder(
```
SebastienGllmt, line 50: we probably want a batch version of this function like `onERC1155BatchReceived` because probably people will want to make a single transaction that creates all the sell orders for all the tokens they have instead of doing it one assetId at a time

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -3,41 +3,59 @@
...
+        uint256 assetId,
+        uint256 assetAmount,
+        uint256 pricePerAsset
+    ) public virtual {
         if (assetAmount == 0 || pricePerAsset == 0) {
             revert InvalidInput(0);
         }
+        asset.safeTransferFrom(msg.sender, address(this), assetId, assetAmount, bytes(""));
```
SebastienGllmt, line 58: Do we want to ask for approval instead of transferring tokens here (I'm not sure which is better)?

We don't need to explicitly transfer anymore after the ERC1155 refactor I think. Previously we had to do it we needed to guarantee the orderID was generated according to a set of rules, but no longer care about orderIDs (instead, we just care about which tokenID is part of an order)

matejos: I've thought about this, but if we are not transferring the tokens to the DEX... Imagine a scenario in which I create a sell order but then I transfer the assets away. How does the DEX know that the sell order is no longer valid? Yes, the fill function would not fill that order (since the transfer would fail), but from the front-end perspective, how do you discard that sell order? You'd have to monitor the asset contract for Transfer events? Doable, but seems kinda tedious. It also does not solve the situations in which the asset gets transferred away just before fill order transaction is mined (and would fail to fill that order), which seems unsolvable, but these situations might be improbable, I admit.
What do you think?

matejos: Personally I think approvals are worth the better UX for sellers (after all, OpenSea works this way), but it does make integrating the front-end more complex. I'm not sure how we rank these aspects on importance.

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
matejos: Added.

SebastienGllmt (approved):

PR #321, opened 2024-03-14
fix: we are storing heights from base chain into parallel chain

ecioppettini:
I think https://github.com/PaimaStudios/paima-engine/pull/314 didn't fix one of the issues properly. With the changes in that PR, we stopped writing to `cde_tracking`  when the base chain has no CDE's, which was the reason the CDE's are processed by network now. But I think now we are inserting block_heights from the base network into the parallel ones during the sync.

This is specially a problem if/when the block heights from both networks overlap, since then the db may fail because of the primary key constraint.

I think this change fixes that, although please check my logic if possible.


SebastienGllmt (approved):
LGTM

PR #323, opened 2024-03-15
erc721 cde: keep track of the address that burned the asset

ecioppettini:
This keeps track of the address that burned an erc721 asset by putting those into a different table.

Potentially this could be just a new row in the existing table, but I think this may be clearer? This way this could potentially be optional.

Note that this removes the burned asset from the existing table. This means it won't show anymore when searching for the assets of the 0 address. This may nor may not be desired, but I'm not really sure.

The main motivation for doing this is that, while this can potentially be done by adding erc721 transfer events to the cde and doing the indexing on the stf (which is also what's suggested by the current erc721 docs, but with the generic cde instead), this doesn't work for burns found in the presync stage, since there is no stf running for those.

### Questions

What's the appropiate utils function for this? I imagine we may want a `getBurnedNfts(cde, address)`?

### DIFF
--- a/packages/engine/paima-sm/src/cde-erc721-transfer.ts
+++ b/packages/engine/paima-sm/src/cde-erc721-transfer.ts
@@ -19,7 +19,7 @@ export default async function processErc721Datum(
   const { to, tokenId } = cdeDatum.payload;
   const toAddr = to.toLowerCase();
 
-  const isBurn = Boolean(toAddr.match(/^0x0+$/g));
+  const isBurn = Boolean(toAddr.toLocaleLowerCase().match(/^0x0+(dead)?$/g));
 
   const updateList: SQLUpdate[] = [];
   try {
@@ -30,6 +30,8 @@ export default async function processErc721Datum(
     const newOwnerData = { cde_id: cdeId, token_id: tokenId, nft_owner: toAddr };
     if (ownerRow.length > 0) {
       if (isBurn) {
+        // we do this to keep track of the owner before the asset is sent to the
+        // burn address
         updateList.push([
           cdeErc721BurnInsert,
           { cde_id: cdeId, token_id: tokenId, nft_owner: ownerRow[0].nft_owner },
--- a/packages/engine/paima-sm/src/cde-erc721-transfer.ts
+++ b/packages/engine/paima-sm/src/cde-erc721-transfer.ts
@@ -32,11 +32,9 @@ export default async function processErc721Datum(
     const newOwnerData = { cde_id: cdeId, token_id: tokenId, nft_owner: toAddr };
     if (ownerRow.length > 0) {
       if (isBurn) {
-        if (cdeDatum.burnScheduledPrefix) {
+        if (cdeDatum.burnScheduledPrefix && !isPresync) {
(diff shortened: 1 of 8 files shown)


packages/engine/paima-sm/src/cde-erc721-transfer.ts
```diff
@@ -13,6 +19,8 @@ export default async function processErc721Datum(
   const { to, tokenId } = cdeDatum.payload;
   const toAddr = to.toLowerCase();
 
+  const isBurn = Boolean(toAddr.match(/^0x0+$/g));
```
SebastienGllmt, line 22: Checking the 0 address isn't enough. You also have to check `0x000000000000000000000000000000000000dEaD`

packages/engine/paima-sm/src/cde-erc721-transfer.ts
```diff
@@ -21,7 +29,15 @@ export default async function processErc721Datum(
     );
     const newOwnerData = { cde_id: cdeId, token_id: tokenId, nft_owner: toAddr };
     if (ownerRow.length > 0) {
-      updateList.push([cdeErc721UpdateOwner, newOwnerData]);
+      if (isBurn) {
+        updateList.push([
```
SebastienGllmt, line 33: Probably good to leave a comment explaining that we're doing this so we can track the address that owned the NFT before it was sent to the burn address

SebastienGllmt (approved):

SebastienGllmt:
> What's the appropiate utils function for this? I imagine we may want a getBurnedNfts(cde, address)?

Don't think this is needed for anything. The main thing we need is a stf update that says who burned which asset

aks the erc721 primitive would be updates so instead of providing only a prefix for the mint event, you can add upbto 2 prefixes: one for the mint and one for the burn

ecioppettini:
> We could build it as a separate primitive like erc20-deposit but add a special shortcut for the burn address

what would be the advantage of that approach? It has the issue that we would end up fetching some data twice, if you have the `erc721` and this new primitive enabled at the same time. But I guess that would be useful if someone only wanted the burn events, is that a potential use-case?

For now I added an optional prefix. If it's not present then the stf events are not emitted.

SebastienGllmt (approved):

packages/engine/paima-sm/src/cde-erc721-transfer.ts
```diff
@@ -21,7 +31,26 @@ export default async function processErc721Datum(
     );
     const newOwnerData = { cde_id: cdeId, token_id: tokenId, nft_owner: toAddr };
     if (ownerRow.length > 0) {
-      updateList.push([cdeErc721UpdateOwner, newOwnerData]);
+      if (isBurn) {
+        if (cdeDatum.burnScheduledPrefix) {
+          const scheduledInputData = `${cdeDatum.burnScheduledPrefix}|${ownerRow[0].nft_owner}|${tokenId}`;
+          const scheduledBlockHeight = isPresync
```
SebastienGllmt, line 37: This behavior for handling presync looks different than what we have in `cde-erc20-deposit`

ecioppettini: Hmmm, right... It's more consistent to not schedule data during the presync at all in this PR.

PR #330, opened 2024-03-25
Inverse Projected 1155 contracts

matejos:


packages/contracts/evm-contracts/contracts/token/InverseAppProjected1155.sol
```diff
@@ -0,0 +1,173 @@
...
+
+    /// @dev Mints `value` of a new token to the transaction sender.
+    /// Increases the `currentTokenId`.
+    /// Reverts if transaction sender is a smart contract that does not implement IERC1155Receiver-onERC1155Received.
+    /// Emits the `Minted` event.
+    /// @param value the amount of tokens to mint.
+    /// @param data additional data to pass to the receiver contract.
+    /// @param verificationData any additional data to verify the validity of the mint
+    /// @return id of the minted token.
+    function mint(
```
SebastienGllmt, line 84: I noticed the mint here is virtual unlike the inverse projected ERC721. Do we want to unify this?

packages/contracts/evm-contracts/contracts/token/IERC4906Agnostic.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+// Forked from OpenZeppelin Contracts (last updated v5.0.0) (interfaces/IERC4906.sol)
+
+pragma solidity ^0.8.13;
+
+/// @title Agnostic Metadata Update Extension
+interface IERC4906Agnostic {
```
SebastienGllmt, line 7: We didn't use this for the inverse projected ERC721. Maybe we should switch to using it there as well?

packages/contracts/evm-contracts/contracts/token/IInverseProjected1155.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {IERC1155MetadataURI} from "@openzeppelin/contracts/token/ERC1155/extensions/IERC1155MetadataURI.sol";
+import {IERC4906Agnostic} from "./IERC4906Agnostic.sol";
+
+/// @dev A standard ERC1155 that can be burned and has a special uri function accepting a custom base URI.
+interface IInverseProjected1155 is IERC1155MetadataURI, IERC4906Agnostic {
```
SebastienGllmt, line 8: Maybe we should add tests for this like we did for the inverse projected ERC721?

packages/contracts/evm-contracts/contracts/token/IERC4906Agnostic.sol
matejos: The reason we use this here is that the OpenZeppelin's implementation of the `IERC4906` interface extends the `IERC721` interface, so we cannot use it but have to have this custom one.

packages/contracts/evm-contracts/contracts/token/IInverseProjected1155.sol
matejos: Definitely, will do

packages/contracts/evm-contracts/contracts/token/InverseAppProjected1155.sol
matejos: Yes! I realized when implementing this that we should basically make all the external functions in the standard contracts virtual, to allow users to override it. I will make that unification in the next PR.

packages/contracts/evm-contracts/contracts/token/InverseBaseProjected1155.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    /// @dev Mints `value` of a new token to transaction sender, passing `initialData` to be emitted in the event.
+    /// Increases the `currentTokenId`.
+    /// Reverts if transaction sender is a smart contract that does not implement IERC1155Receiver-onERC1155Received.
+    /// Emits the `Minted` event.
+    /// Returns the id of the minted token.
+    function mint(
+        uint256 value,
+        bytes memory data,
+        string calldata initialData
```
SebastienGllmt, line 59: although it's clear the difference between `data` and `initialData` if you're familiar with both ERC1155 and Paima, probably we should mention something in the comment since it may not be obvious for others (I noticed we put a comment to explain this in the app-initiated variation, but not this one)

packages/contracts/evm-contracts/contracts/token/IUri.sol
```diff
@@ -0,0 +1,13 @@
...
+
+/// @dev An interface exposing the `uri` function from IERC1155MetadataURI.
+interface IUri {
+    /**
+     * @dev Returns the URI for token type `id`.
+     *
+     * If the `\{id\}` substring is present in the URI, it must be replaced by
+     * clients with the actual token type ID.
+     */
+    function uri(uint256 id) external view returns (string memory);
```
SebastienGllmt, line 12: What's the reason for using this instead of `IERC1155MetadataURI` directly?

matejos: `IERC1155MetadataURI` inherits `IERC1155`, so using it would dictate all requirements of `IERC1155`, which are not needed in `uri(uint256 id, IUri customUriInterface)` function call.

packages/contracts/evm-contracts/contracts/token/InverseBaseProjected1155.sol
matejos: comment added

SebastienGllmt (approved):

Issue #325, opened 2024-03-21
`[cardano-transfer primitive]` returns more than transfers


SebastienGllmt:
Currently, the `cardano-transfer` primitive queries the user-provided credential in `extensions.yml` directly to the Carp history endpoint.

However, the Carp history endpoint looks for anywhere this credential may have appeared in *any* structure in Cardano (not just transfers).

As a reminder, there are 2 main use-cases where you might want to track an address:
1. Output for when you want to check when somebody sends money to an address (ex: buying an in-game asset by sending money to an address)
2. Input for when you want to check when somebody executes a specific smart contract (since Cardano smart contracts appear in the input)

The filter we want to use depends on the situation:
1. For outputs, we need just `RelationFilterType.Output`
2. For smart contract inputs, we need
    1. `RelationFilterType.Input`
    2. `RelationFilterType.StakeRegistration`
    3. `RelationFilterType.StakeDeregistration`
    4. `RelationFilterType.StakeDelegation`
    5. `RelationFilterType.Withdrawal`

This should fix the problem. We should be explicit in the docs though that this means that this primitive will **not** match on transactions where just the staking key matches the credential and not the payment key. I think this is fine because checking if somebody sent to an address with a specific staking key seems like an anti-pattern and they should probably instead be using the `cardano-stake-delegation` primitive

ecioppettini:
> The filter we want to use depends on the situation:

Do we want this to be a setting? Or would https://github.com/PaimaStudios/paima-engine/pull/332 be enough?

PR #332, opened 2024-03-25
cardano transfer: narrow credential relation filter

ecioppettini:
implements #325

### DIFF
--- a/packages/engine/paima-funnel/src/cde/cardanoTransfer.ts
+++ b/packages/engine/paima-funnel/src/cde/cardanoTransfer.ts
@@ -25,13 +25,7 @@ export default async function getCdeData(
 ): Promise<CdeCardanoTransferDatum[]> {
   let result = [] as CdeCardanoTransferDatum[];
 
-  const relationFilter =
-    RelationFilterType.Output |
-    RelationFilterType.Input |
-    RelationFilterType.StakeRegistration |
-    RelationFilterType.StakeDeregistration |
-    RelationFilterType.StakeDelegation |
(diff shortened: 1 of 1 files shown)


packages/engine/paima-funnel/src/cde/cardanoTransfer.ts
```diff
@@ -24,10 +25,18 @@ export default async function getCdeData(
 ): Promise<CdeCardanoTransferDatum[]> {
   let result = [] as CdeCardanoTransferDatum[];
 
+  const relationFilter =
+    RelationFilterType.Output |
+    RelationFilterType.Input |
+    RelationFilterType.StakeRegistration |
+    RelationFilterType.StakeDeregistration |
+    RelationFilterType.StakeDelegation |
+    RelationFilterType.Withdrawal;
```
SebastienGllmt, line 34: Hmm, this way is possible I guess, but we'd have to be clear in the docs that users have to manually parse the transaction to know if the address appears in the inputs or outputs (and if it appears in neither the inputs or the outputs, it's because it's one of the other types)

SebastienGllmt: I can't think of a particular nice way to do this other than to include the tx cbor as well (as mentioned originally here: https://github.com/PaimaStudios/paima-engine/issues/262)

SebastienGllmt: Also I just realized we also probably want `RequiredSigner` as an input type

PR #333, opened 2024-03-26
add i/o addresses to cardano mint and burn

ecioppettini:
For some reason I can't get the temporal file dependency to carp-client work anymore. But this seems to work otherwise.


SebastienGllmt (approved):
Approving modulo the need to replace the package.json

PR #334, opened 2024-03-26
Feature/delegate wallet passthrough

acedward:
```
        delegate                = &wd|from?|to?|from_signature|to_signature
        migrate                 = &wm|from?|to?|from_signature|to_signature
        cancelDelegate          = &wc|to_signature
```
if valid wallet-delegate commands are now passed to the SFT.
Valid means the signatures are correct, and the command could be processed and was not rejected.

### DIFF
--- a/packages/engine/paima-sm/src/index.ts
+++ b/packages/engine/paima-sm/src/index.ts
@@ -417,50 +417,50 @@ async function processUserInputs(
     // cancelDelegate = &wc|to_signature
     const delegateWallet = new DelegateWallet(DBConn);
     if (inputData.inputData.startsWith(DelegateWallet.INTERNAL_COMMAND_PREFIX)) {
-      await delegateWallet.process(
+      const status = await delegateWallet.process(
         inputData.realAddress,
         inputData.userAddress,
         inputData.inputData
       );
-    } else {
+      if (!status) continue;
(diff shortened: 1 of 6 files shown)


packages/engine/paima-sm/src/index.ts
```diff
@@ -417,50 +417,50 @@ async function processUserInputs(
...
+      inputData.userId = newAddress.id;
+    }
+
+    // Trigger STF
+    const sqlQueries = await gameStateTransition(
+      inputData,
+      latestChainData.blockNumber,
+      randomnessGenerator,
+      DBConn
+    );
```
SebastienGllmt, line 438: What happens if the user's STF doesn't have something to parse this? What happens if the name conflicts with something else the user defined in their STF

acedward: 1. If the user does not have the parser setup, a console error will show up, but not halt paima-engine. (same error if a user posts any invalid command directly to the blockchain)

2. Conflict is unlikely as the command starts with the symbol `&` as described in the documentation.

image

SebastienGllmt: Okay, this may be a bit messy until we make STF failure more local instead of failing full txs, but we don't need to block this PR on that

SebastienGllmt (approved):

PR #337, opened 2024-04-02
Mint without verification data should not make an external call

matejos:


packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
```diff
@@ -61,7 +61,7 @@ contract InverseAppProjectedNft is IInverseAppProjectedNft, ERC721, Ownable {
             super.supportsInterface(interfaceId);
     }
 
-    function currentNonce(address user) external view returns (uint256) {
+    function currentNonce(address user) public view returns (uint256) {
```
SebastienGllmt, line 64: This no longer matches the interface

packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
```diff
@@ -75,7 +75,7 @@ contract InverseAppProjectedNft is IInverseAppProjectedNft, ERC721, Ownable {
     /// Increases the `totalSupply` and `currentTokenId`.
     /// Reverts if `_to` is a zero address or if it refers to smart contract but does not implement IERC721Receiver-onERC721Received.
     /// Emits the `Minted` event.
-    function mint(address _to, bytes calldata _verificationData) public virtual returns (uint256) {
+    function mint(address _to, bytes memory _verificationData) public virtual returns (uint256) {
```
SebastienGllmt, line 78: This no longer matches the interface

packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
```diff
@@ -95,8 +95,8 @@ contract InverseAppProjectedNft is IInverseAppProjectedNft, ERC721, Ownable {
         return tokenId;
     }
 
-    function mint(address _to) external returns (uint256) {
-        return this.mint(_to, "");
+    function mint(address _to) public returns (uint256) {
```
SebastienGllmt, line 98: This no longer matches the interface

packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
matejos: Not true, functions in interfaces **must** be declared external, implementation can be declared public.

packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
matejos: Not true, functions in interfaces **must** be declared external, implementation can be declared public.

packages/contracts/evm-contracts/contracts/token/InverseAppProjectedNft.sol
matejos: Not true, quite the contrary, it **now** matches the interface

PR #338, opened 2024-04-04
Fix batcher crashing silently

SpaceManiac:
If the batcher crashed during initialization, for example because of an invalid configuration such that `getRemoteBackendVersion` failed, no error would be printed and the batcher would exit with status code 0. Instead, print that error and exit with an error code.

Such errors were previously being suppressed by the `on('uncaughtException')` handler, because they occurred within 1 second of startup.

### DIFF
--- a/packages/batcher/runtime/src/index.ts
+++ b/packages/batcher/runtime/src/index.ts
@@ -190,6 +190,9 @@ async function main(): Promise<void> {
 }
 
 main().catch(e => {
+  // we catch the error here instead of relying on `uncaughtException` monitoring
+  // because if an exception causes us to reach this line, it means the error was thrown during initalization
+  // which is typically something we cannot recover from
   console.error(e);
   process.exit(1);
 });
--- a/packages/batcher/runtime/src/index.ts
+++ b/packages/batcher/runtime/src/index.ts
@@ -189,4 +189,7 @@ async function main(): Promise<void> {
   await runtime.run(gameInputValidator, batchedTransactionPoster, provider);
 }
 
-void main();
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
+});


packages/batcher/runtime/src/index.ts
```diff
@@ -189,4 +189,7 @@ async function main(): Promise<void> {
   await runtime.run(gameInputValidator, batchedTransactionPoster, provider);
 }
 
-void main();
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
```
SebastienGllmt, line 194: shouldn't this be caught under the `process.on('uncaughtException'` we have at the top of this file?

SpaceManiac: Ah, I bet the problem is that an error during startup is still within 1000ms of the starting `lastCatchDate`, that's why it's being swallowed entirely.

Fixing that still gets us an error message + exit, although it's with the 0 success code. It is a simpler fix if you don't care about that.

```diff
-let lastCatchDate = Date.now();
+let lastCatchDate = Date.now() - MINIMUM_INTER_CATCH_PERIOD;
```

packages/batcher/runtime/src/index.ts
```diff
@@ -189,4 +189,7 @@ async function main(): Promise<void> {
   await runtime.run(gameInputValidator, batchedTransactionPoster, provider);
 }
 
-void main();
+main().catch(e => {
+  console.error(e);
```
SebastienGllmt, line 193: At the least I would add a comment like this

```suggestion
  // we catch the error here instead of relying on `uncaughtException` monitoring
  // this is because if an exception causes us to reach this line, it means the error was thrown during initalization
  // which is typically something we cannot recover from
  console.error(e);
```

If we want to make this a bit more obvious, we could split up the `main()` function into 2 parts:
1. An initialize function
2. The last 2 lines (`requestStart` and `runtime.run`)

then we could add the try/catch just on the `initialize` function

I'm okay with either approach

SebastienGllmt (approved):

PR #347, opened 2024-04-12
Fix ssh-agent requirement by using https to clone docs

SpaceManiac:
The ssh-agent is only needed because the build process uses SSH to clone GitHub repos. Because these are public repos, we can use HTTPS instead of SSH and bypass this requirement entirely. Now the build will hang neither at the beginning nor the end.

Also `cp -f` will allow the `paima-engine-linux` binary to be replaced on the filesystem even if it's currently running (usually not an issue, I assume Node mmaps it or something).


SebastienGllmt (approved):
Yeah we used to copy private repos and I've been trying to slowly inline everything into this monorepo. I'd like to get the docs in this repo someday to avoid the current awkward work to try and make the docs update at the same time as the code and be properly versioned. Maybe someday

PR #348, opened 2024-04-15
Add ERC-1155 funnel primitive

SpaceManiac:
Based somewhat on the ERC-20 and ERC-721 primitives. Works with any IERC1155 contract, not specific to IInverseAppProjected1155. Utility functions provided to read the `cde_erc1155_data` table that tracks current balances based on transfer history.

Example configuration:
```yml
extensions:
  - name: "Whatever"
    type: "erc1155"
    contractAddress: "..."
    startBlockHeight: 0
    scheduledPrefix: "my1155_transfer"  # optional
    burnScheduledPrefix: "my1155_burn"  # optional
    network: "..."
```

Scheduled event structure:
```
scheduledPrefix|0xOperator|0xFrom|0xTo|["id0","id1",...]|["amount0","amount1",...]
burnScheduledPrefix|0xOperator|0xFrom|["id0","id1",...]|["amount0","amount1",...]
```

Mints are from=`0x0...0`, burns are to=`0x0...0` or `0x0...dead`.

---

Open questions:
* [x] Is `contractAddress` actually useful to include in the events?
  * Eddie says no, so TODO: remove it
* [x] Does a chain ID need to be mixed in here somehow?
  * Eddie says no
* [x] I see for ERC-721 we autodetect whether the contract is base IERC1155 or child IInverseBaseProjectedNft. Should we do the same here? But it might be confusing when sometimes the "paima" version is BaseProjected and sometimes the "paima" version is AppProjected.
  * The important event is transfers. IInverseAppProjected1155 has a special "Minted" event but Tarochi at least doesn't need it and it can be observed either generically or through a future extension.
  * [x] TODO: clean up leftovers of IInverseAppProjected1155
* [x] Is the blockchain -> CDE datum step the right place to normalize single and batch transfers, or should this be pushed later, possibly to the app?
  * Eddie suggests normalizing to batch instead of to single
* [x] Should there be some attempt for ERC-20 and ERC-721's utility functions to transparently support ERC-1155, or should ERC-1155 just get its own suite?
  * [x] TODO: write these


packages/engine/paima-sm/src/types.ts
```diff
@@ -249,6 +282,7 @@ export enum CdeEntryTypeName {
   ERC20Deposit = 'erc20-deposit',
   ERC721 = 'erc721',
   ERC6551Registry = 'erc6551-registry',
+  InverseAppProjected1155 = 'erc1155-app',
```
acedward, line 285: Any reason to call `erc1155-app` instead of just `erc1155` ?

SpaceManiac: I'm imagining that the IInverseBaseProjected1155 contract might be used later and could then be named `erc1155-base`, but maybe that's something that (like `erc721`) can be detected from the contract on startup.

packages/engine/paima-funnel/src/cde/inverseAppProjected1155.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function transferBatchToDatums(
+  event: TransferBatch,
+  extension: ChainDataExtensionInverseAppProjected1155,
+  network: string
+): CdeInverseAppProjected1155TransferDatum[] {
+  // NOTE: assumes that ids and values have the same length
+  const result: CdeInverseAppProjected1155TransferDatum[] = [];
+  for (let i = 0; i < event.returnValues.ids.length; ++i) {
```
acedward, line 101: I feel that's better to send one payload to the app, than many small ones.

Why? E.g., if I buy Gold in the Store, a single purchase might have to transfer many TokenIds content in a batchTransfer.

So from the game's point of view, only one transfer was done ( I have to show them 1 notification " Player Z has purchased 10000 Gold " ) instead of many small transfers, that might have been done to comply with the sale.

packages/engine/paima-sm/src/cde-erc1155-transfer.ts
```diff
@@ -0,0 +1,18 @@
...
+export default async function processInverseAppProjected1155TransferDatum(
+  cdeDatum: CdeInverseAppProjected1155TransferDatum,
+  inPresync: boolean
+): Promise<SQLUpdate[]> {
+  const { transferScheduledPrefix, contractAddress, payload, blockNumber } = cdeDatum;
+  if (!transferScheduledPrefix || inPresync) {
+    return [];
+  }
+  const { operator, from, to, id, value } = payload;
+  const scheduledBlockHeight = Math.max(blockNumber, ENV.SM_START_BLOCKHEIGHT + 1);
```
ecioppettini, line 15: same comment as above

packages/engine/paima-sm/src/cde-erc1155-mint.ts
```diff
@@ -0,0 +1,18 @@
...
+export default async function processInverseAppProjected1155MintDatum(
+  cdeDatum: CdeInverseAppProjected1155MintDatum,
+  inPresync: boolean
+): Promise<SQLUpdate[]> {
+  const { mintScheduledPrefix, contractAddress, payload, blockNumber } = cdeDatum;
+  if (!mintScheduledPrefix || inPresync) {
+    return [];
+  }
+  const { tokenId, minter, userTokenId, value } = payload;
+  const scheduledBlockHeight = Math.max(blockNumber, ENV.SM_START_BLOCKHEIGHT + 1);
```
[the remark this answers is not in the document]
SpaceManiac: Got it. I'll make the change if that PR is merged first

packages/engine/paima-sm/src/types.ts
SpaceManiac: The most important event is transfers which are available on all ERC-1155, not specific to IInverseAppProjected1155, so in supporting that I've changed this to `erc1155` too.

packages/node-sdk/paima-utils-backend/src/cde-access.ts
```diff
@@ -15,8 +15,18 @@ import {
...
+  internalGetErc1155ByTokenId,
+  internalGetErc1155ByTokenIdAndWallet,
 } from './cde-access-internals.js';
-import type { ICdeCardanoGetProjectedNftResult } from '@paima/db/src';
+import type {
+  ICdeCardanoGetProjectedNftResult,
+  ICdeErc1155GetAllTokensResult,
+  ICdeErc1155GetByTokenIdAndWalletResult,
+  ICdeErc1155GetByTokenIdResult,
+} from '@paima/db/src';
```
SebastienGllmt, line 28: don't import from `src` directory. If you import `src` folders TS is not smart enough to know you're trying to leverage the build output of the `@paima/db` compilation unit, and it will instead inline the src folders part of this compilation unit instead.

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -158,6 +167,50 @@ export async function internalGetGenericDataBlockheightRange(
...
+  readonlyDBConn: Pool,
+  cde_id: number,
+  wallet_address: string
+): Promise<bigint> {
+  const results = await cdeErc1155GetTotalBalanceAllTokens.run(
+    { cde_id, wallet_address },
+    readonlyDBConn
+  );
+  return BigInt(results[0].total ?? '0');
+}
```
SebastienGllmt, line 180: Isn't this dangerous as a utility function?

1. For ERC-1155 that are not for inverse projections, this utility function is never useful (since nobody else has erc1155 collections like us where all the tokens are the same)
2. For inverse projection use-cases, we need extra information to know which token ID is a valid mint and which one isn't

packages/engine/paima-runtime/src/cde-config/utils.ts
```diff
@@ -25,13 +25,13 @@ export function getEarliestStartSlot(config: ChainDataExtension[]): number {
 }
 
 // returns pair [rawAbiFileData, artifactObject.abi]
```
SebastienGllmt, line 27: This comment is no longer valid because you changed the return type

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
SpaceManiac: Hmm, makes sense. Previously the Tarochi side was doing this query, but it never inserted invalid mints to begin with. Now Tarochi instead cross-references with its table of which mints were valid. I've gone ahead and trashed this query.

SebastienGllmt (approved):

PR #351, opened 2024-04-22
Fix mismatched HTML tags

findnature:


SebastienGllmt:
Closing since I assume this is just airdrop farming

PR #353, opened 2024-04-22
Add Dockerfile for engine, batcher, and SDK images

SpaceManiac:
A bit of an experiment. The goal is to make it easier to version and deploy games, and to reduce accidental structural differences between dev, test, and production.

`./docker-tag [tag]` builds and tags the images with the given tag (ex: `2.3.0`) or `latest` if not specified.

The intended use of the images are:

* `paima-engine` serves as a base for game backend images:
  ```Dockerfile
  FROM ghcr.io/paimastudios/paima-engine:2.3.0 AS mygame-backend
  COPY --from=build-backend /src/packaged packaged
  ARG NETWORK=docker
  COPY .env.$NETWORK .env.$NETWORK
  COPY config.$NETWORK.yml config.$NETWORK.yml
  COPY extensions.$NETWORK.yml extensions.yml
  ENV NETWORK=$NETWORK
  ```

* `paima-batcher` can be run directly, supplying just the configuration:
  ```sh
  docker run --env-file .env.docker ghcr.io/paimastudios/paima-batcher:2.3.0
  ```
  or included in a docker-compose file:
  ```yaml
  batcher:
    image: ghcr.io/paimastudios/paima-batcher:2.3.0
    depends_on:
      - hardhat
      - backend
      - database
    env_file:
      - .env.docker
    ports:
      - "3340:3340"
  ```

* `paima-sdk` can be used as a `COPY` source for a Docker build that needs a newer Node SDK than is on NPM:
  ```Dockerfile
  COPY --from=ghcr.io/paimastudios/paima-sdk:latest \
    /node_modules/@paima \
    node_modules/@paima
  ```

Caveats:
- The `paima-engine` image is bigger than it needs to be because the binary includes the batcher, contracts, templates, so on. The image only really needs the `run` subcommand.
- The old batcher and db docker-compose templates are still hanging around.
- The `paima-sdk` image is a little goofy. An image of just free-hanging files is not really how Docker is meant to be used. But this seems like the easiest way to make compiled outputs of non-NPM-tagged versions available to downstream Docker builds.

Thoughts?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,36 @@
+# NOTE: unlike .gitignore, ** is not implied, and only the root .dockerignore is considered.
+
+# Dockerignore exclusive
+/.dockerignore
+/Dockerfile
+/docker-tag
+/.git
+
+# .gitignore
+/**/node_modules
+/**/bin
+/**/.nx
+/**/dist
+/**/build
+/**/*.tsbuildinfo
+/**/*.log
+/**/.idea
+
+# .husky/_/.gitignore
+/.husky/_
+
+# packages/batcher/batcher-standalone/.gitignore
+/packages/batcher/batcher-standalone/packaged
+
+# packages/contracts/evm-contracts/.gitignore
+/packages/contracts/evm-contracts/artifacts
+/packages/contracts/evm-contracts/build
+/packages/contracts/evm-contracts/cache
+/packages/contracts/evm-contracts/publish
+/packages/contracts/evm-contracts/typechain-types
+
+# packages/engine/paima-standalone/.gitignore
+/packages/engine/paima-standalone/packaged
+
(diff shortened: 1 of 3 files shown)


acedward:
I just tested it and failed for me, I had to add `python3 g++ make` to the Dockerfile

image

image

image

SpaceManiac:
Interesting. I see "gyp info using node@20.8.1 | linux | arm64" in the logs. I think since you're on an M1 Mac, it's building a Linux-for-arm image by default, while I'm on x86_64. We could set `--platform linux/amd64` in either the `docker-tag` script or in the `Dockerfile` to override that default.

acedward:
We moved to Effectstream and this is not compatible.
We should perhaps deploy the engine as base image.

PR #355, opened 2024-04-25
mina/auro wallet support

ecioppettini:


packages/paima-sdk/paima-crypto/src/mina.ts
```diff
@@ -0,0 +1,42 @@
+import { doLog } from '@paima/utils';
+import type { IVerify } from './IVerify.js';
+
+export class MinaCrypto implements IVerify {
+  verifyAddress = async (address: string): Promise<boolean> => {
+    // base58 alphabet
+    return await Promise.resolve(
+      /^[1|2|3|4|5|6|7|8|9|A|B|C|D|E|F|G|H|J|K|L|M|N|P|Q|R|S|T|U|V|W|X|Y|Z|a|b|c|d|e|f|g|h|i|j|k|m|n|o|p|q|r|s|t|u|v|w|x|y|z]/.test(
```
SebastienGllmt, line 8: This is only checking the first letter in the string. Also, does Mina use regular base58 or checked base58 (with a checksum)?

Probably it's easier to check the address validity with regular JS instead of regex (esp. if it's using the checksum variant)

packages/paima-sdk/paima-mw-core/src/errors.ts
```diff
@@ -57,6 +57,7 @@ export const enum PaimaMiddlewareErrorCode {
   INTERNAL_INVALID_POSTING_MODE,
   // only to be used as a counter
   FINAL_PAIMA_GENERIC_ERROR = 1_000_000,
+  MINA_LOGIN,
```
SebastienGllmt, line 60: Nothing should come after `FINAL_PAIMA_GENERIC_ERROR`. Probably you meant to put this after `ALGORAND_LOGIN`

The reason it's setup this way is because enums are (by default) incrementing numbers. Therefore, unless special care is taken, adding any new value to an enum is a breaking change (it shifts the value of everything that comes after it by one). To avoid this, we've properly namespaced all the enums (which is why you see some of them have explicit values), and `FINAL_PAIMA_GENERIC_ERROR` comes at the end so the users know that everything after this is a user-space error (and not a Paima Engine error)

packages/paima-sdk/paima-providers/src/mina.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    if ('signature' in signed) {
+      const { signature } = signed;
+
+      const networkKind = await this.conn.api.requestNetwork();
+
+      // only these two options are used
+      // https://github.com/aurowallet/auro-wallet-browser-extension/blob/624de322dd99baaa09617bbad4a4a838f0a88edc/src/background/lib/index.js#L17
+      // this is part of the hashed message
+      const network = networkKind.chainId === 'mainnet' ? 'mainnet' : 'testnet';
```
SebastienGllmt, line 147: This is not true. Zeko is being added to Auro (there will be a canonical Zeko L2). Later on, there will probably be many Zeko-based networks in Auro - especially because for Paima we will probably end up where many games run as L3s

packages/paima-sdk/paima-providers/src/mina.ts
```diff
@@ -0,0 +1,169 @@
...
+    } else {
+      throw new ProviderApiError('Failed to sign message');
+    }
+  };
+
+  static fetchAddress = async (conn: ActiveConnection<MinaApi>): Promise<string> => {
+    try {
+      const addresses = await conn.api.requestAccounts();
+      if ('length' in addresses && addresses.length > 0) {
+        return addresses[0];
```
SebastienGllmt, line 159: Is it okay to just take `addresses[0]` here any ignore the other ones? Does this mean we don't allow people to connect accounts other than their first one to Paima Engine?

packages/paima-sdk/paima-providers/src/mina.ts
ecioppettini: Well... But it's still not the case right now, right? The problem is that the `signMessage` api doesn't tell you what's used for signing, and that's needed to verify. So to have this working for zeko I would need to know what will be the `chainId` in that case, and what will be the value used in there. Either that or the batcher needs to be pre-configured with the network id.

SebastienGllmt: Okay I asked about this and for now Zeko just uses "testnet" so it should work as-is for now with the code you have. I sent a message to the Auro team to see if they have any public documentation for how they plan to change these endpoints going forward, but it shouldn't block us for this PR. I created an issue to make sure we don't forget about this: https://github.com/PaimaStudios/paima-engine/issues/357

packages/paima-sdk/paima-providers/src/mina.ts
SebastienGllmt: We do have `getWindow()?.ethereum?.on('accountsChanged'` for EVM

We can create an issue to implement some kind of listener to Mina in the future to avoid blocking this PR

ecioppettini: You're right. I missed that one for some reason. Then I guess this one should work the same way.

There is the same event on the auro api: https://docs.aurowallet.com/general/reference/api-reference/events#accountschanged

I can add that same kind of code here.

packages/paima-sdk/paima-crypto/src/mina.ts
ecioppettini: good catch... It's checksummed I think: https://github.com/o1-labs/o1js/blob/3b3bd1020a032b9a18d4d97ca445ca4c50b93cfd/src/lib/util/base58.ts#L36

But I'm not sure what can I use to access that function. I could add the `o1js` dependency I guess, and use `PublicKey.fromBech58` or something like that I guess? Because I think that module is not exposed.

SebastienGllmt: I assume their implementation is just `base58Check` and there are npm packages that provide just this functionality (not sure why o1js inlined it into their code instead of just using one of these packages)

packages/paima-sdk/paima-providers/src/mina.ts
ecioppettini: I added the same kind of handler now. But to be honest I'm not sure who it's supposed to work. Since the invalidation of the provider is in the connector and not on the provider itself, but I think the connection is just created once anyway, so I'm not really sure how is it supposed to be used.

SebastienGllmt (approved):

PR #361, opened 2024-05-03
add command to print the json schema of the config

ecioppettini:
I think it's nice to have some intellisense when editing the config, and we already have the json schemas for free with typebox.

I want this to use it with [yaml-language-server](https://github.com/redhat-developer/yaml-language-server) like this:

```
./paima-engine-linux gen-config-json-schema | jq > config-json-schema.json
```

And then in `config.localhost.yml`

```
# yaml-language-server: $schema=config-json-schema.json

Network:
  type: evm-main
  chainUri: 'http://localhost:8545'
  chainId: 31337
  chainCurrencyName: 'ETH'
  chainCurrencySymbol: 'ETH'
  chainCurrencyDecimals: 18
  blockTime: 2
  paimaL2Contract: ...
```

### DIFF
--- a/packages/engine/paima-standalone/src/utils/input.ts
+++ b/packages/engine/paima-standalone/src/utils/input.ts
@@ -17,6 +17,7 @@ import {
 import { importOpenApiJson, importTsoaFunction } from './import.js';
 import type { Template } from './types.js';
 import RegisterRoutes, { EngineService } from '@paima/rest';
+import { BaseConfigWithoutDefaults } from '@paima/utils/src/config/loading.js';
 
 // Prompt user for input in the CLI
 export const userPrompt = (query: string): Promise<string> => {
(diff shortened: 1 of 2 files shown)


packages/engine/paima-standalone/src/utils/input.ts
```diff
@@ -17,6 +17,7 @@ import {
 import { importOpenApiJson, importTsoaFunction } from './import.js';
 import type { Template } from './types.js';
 import RegisterRoutes, { EngineService } from '@paima/rest';
+import { BaseConfigWithoutDefaults } from '@paima/utils/src/config/loading.js';
```
SebastienGllmt, line 20: don't use this kind of import style (referencing a .js file from an NPM path). It can cause issues with compilation units

SebastienGllmt (approved):
Other than my small comment, LGTM. We might want to consider adding this to all the templates so I created an issue to track this idea: https://github.com/PaimaStudios/paima-game-templates/issues/70

PR #362, opened 2024-05-09
setup for dynamic evm extensions

ecioppettini:
# About

This adds a new primitive, which is used to instantiate new erc721 primitives at runtime. I tried to keep the code open for extension to other target primitives, but keeping it to erc721 for now to simplify test cases. Currently it should work for both the block funnel and the parallel evm funnel.

### Example
```yaml
# extensions.yaml
extensions:
  - name: "Dynamic erc721"
    type: "dynamic-primitive"
    startBlockHeight: 0
    network: "Hardhat"
    contractAddress: "0x5FbDB2315678afecb367f032d93F642f64180aa3"
    scheduledPrefix: "derc721"
    eventSignature: "MyEvent(address)"
    abiPath: "../dynamic-contract/out/EventEmitter.sol/EventEmitter.json"
    targetType: "erc721"
    fields:
      contractAddress: value
```

The primitive is based on the generic one, so it also uses a generic `eventSignature`, and it requires an abi to be provided in `abiPath`. This is because there is no standard interface for this otherwise, so I tried to keep it somewhat flexible.

- `scheduledPrefix` is the prefix _of the generated primitives_ . There is no scheduled data for the `dynamic-primitive`. This is the same for each one of the generated primitives, since I think otherwise it's not possible to write the parsing rules (if it depended on something dynamic).
- `targetType` can only be `erc721` for now.
- `fields` is used to map the event field name from the contract. So in this example the event has the following signature.

```solidity
event MyEvent(address indexed value);
```

This is used to populate the `contractAddress` field in the resulting erc721 configuration.

### DIFF
--- a/packages/engine/paima-funnel/src/funnels/emulated/utils.ts
+++ b/packages/engine/paima-funnel/src/funnels/emulated/utils.ts
@@ -54,8 +54,8 @@ export function emulateCde(
     .flat()
     .sort((a, b) => {
       const bhDiff = a.blockNumber - b.blockNumber;
-      const cdeIdDiff = a.cdeId - b.cdeId;
-      return bhDiff ? bhDiff : cdeIdDiff;
+      const cdeNameDiff = a.cdeName.localeCompare(b.cdeName);
+      return bhDiff ? bhDiff : cdeNameDiff;
     })
     .map(cdeDatum => ({ ...cdeDatum, blockNumber }));
 }
--- a/packages/engine/paima-funnel/src/reading.ts
+++ b/packages/engine/paima-funnel/src/reading.ts
@@ -8,6 +8,7 @@ import {
   doLog,
   ChainDataExtensionType,
   getErc721Contract,
+  DYNAMIC_PRIMITIVE_NAME_SEPARATOR,
 } from '@paima/utils';
 import type { PaimaL2Contract } from '@paima/utils';
 import { TimeoutError } from '@paima/runtime';
@@ -180,7 +181,7 @@ export async function fetchDynamicEvmPrimitives(
     for (const _ext of exts) {
       const ext: CdeDynamicEvmPrimitiveDatum = _ext as CdeDynamicEvmPrimitiveDatum;
 
-      const cdeName = `${ext.cdeName}-${
+      const cdeName = `${ext.cdeName}${DYNAMIC_PRIMITIVE_NAME_SEPARATOR}${
         sharedData.extensions.filter(
           e =>
             e.cdeType !== ChainDataExtensionType.DynamicEvmPrimitive &&
--- a/packages/engine/paima-funnel/src/reading.ts
+++ b/packages/engine/paima-funnel/src/reading.ts
@@ -13,7 +13,7 @@ import type { PaimaL2Contract } from '@paima/utils';
 import { TimeoutError } from '@paima/runtime';
 import {
   CdeEntryTypeName,
-  type CdeDynamicPrimitiveDatum,
+  type CdeDynamicEvmPrimitiveDatum,
   type ChainData,
   type ChainDataExtensionDatum,
 } from '@paima/sm';
@@ -155,28 +155,28 @@ async function getPaimaEvents(
 // change the set of primitives. Otherwise we would miss the events for those.
 // If there is an event that triggers a dynamic primitive here, then an
 // extension will be added to the list with a starting block at that point.
-export async function fetchDynamicPrimitives(
+export async function fetchDynamicEvmPrimitives(
   fromBlock: number,
   toBlock: number,
   web3: Web3,
   sharedData: FunnelSharedData,
   network: string
 ): Promise<ChainDataExtensionDatum[][]> {
-  const dynamicPrimitives = await getUngroupedCdeData(
+  const DynamicEvmPrimitives = await getUngroupedCdeData(
     web3,
     sharedData.extensions.filter(
       extension =>
         extension.network === network &&
-        extension.cdeType === ChainDataExtensionType.DynamicPrimitive
+        extension.cdeType === ChainDataExtensionType.DynamicEvmPrimitive
     ),
     fromBlock,
     toBlock,
     network
   );
 
-  for (const exts of dynamicPrimitives) {
+  for (const exts of DynamicEvmPrimitives) {
     for (const _ext of exts) {
-      const ext: CdeDynamicPrimitiveDatum = _ext as CdeDynamicPrimitiveDatum;
+      const ext: CdeDynamicEvmPrimitiveDatum = _ext as CdeDynamicEvmPrimitiveDatum;
 
       // this would propagate the change to further funnels in the pipeline,
       // which is needed to set the proper cdeId.
@@ -197,5 +197,5 @@ export async function fetchDynamicPrimitives(
     }
   }
 
-  return dynamicPrimitives;
+  return DynamicEvmPrimitives;
 }
--- a/packages/engine/paima-funnel/src/reading.ts
+++ b/packages/engine/paima-funnel/src/reading.ts
@@ -1,13 +1,27 @@
 import type Web3 from 'web3';
 import type { BlockTransactionString } from 'web3-eth';
 import type { PoolClient } from 'pg';
-import { timeout, cutAfterFirstRejected, DEFAULT_FUNNEL_TIMEOUT, doLog } from '@paima/utils';
+import {
+  timeout,
+  cutAfterFirstRejected,
+  DEFAULT_FUNNEL_TIMEOUT,
+  doLog,
+  ChainDataExtensionType,
+  getErc721Contract,
+} from '@paima/utils';
 import type { PaimaL2Contract } from '@paima/utils';
 import { TimeoutError } from '@paima/runtime';
-import type { ChainData } from '@paima/sm';
+import {
+  CdeEntryTypeName,
+  type CdeDynamicPrimitiveDatum,
+  type ChainData,
+  type ChainDataExtensionDatum,
+} from '@paima/sm';
 import type { PaimaGameInteraction } from '@paima/utils';
 
 import { extractSubmittedData } from './paima-l2-processing.js';
+import type { FunnelSharedData } from './funnels/BaseFunnel.js';
+import { getUngroupedCdeData } from './cde/reading.js';
 
 export async function getBaseChainDataMulti(
   web3: Web3,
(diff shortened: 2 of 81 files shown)


packages/engine/paima-sm/src/types.ts
```diff
@@ -521,6 +536,32 @@ export type ChainDataExtensionMinaActionGeneric = ChainDataExtensionBase &
     cdeType: ChainDataExtensionType.MinaActionGeneric;
   };
 
+export const ChainDataExtensionDynamicPrimitiveConfig = Type.Intersect([
+  ChainDataExtensionConfigBase,
+  Type.Object({
+    type: Type.Literal(CdeEntryTypeName.DynamicPrimitive),
+    targetType: Type.Literal(CdeEntryTypeName.ERC721),
```
SebastienGllmt, line 543: I'm not sure I understand why we have a target type instead of making the dynamic primitive "wrap" another arbitrary primitive type by nesting the YML config or something similar

ecioppettini: Well, my assumption was that you can't add the full config since some of the data may come in the event (like now with contract address). But maybe we can wrap the yml config making everything optional? The idea then would be to have a partial config here and then have a way to fill the rest from the contract event.

SebastienGllmt: Yeah I think the current setup you have with a dynamic NFT that fills in based on a `fields` map makes a lot of sense. It's not as clear to me:
1. Why we hardcoded this to only ERC721 (it feels like everything should work with other standards as-is)
2. The fact that it wraps is not immediate obvious from the shape of the configuration

I think your idea of making all the fields in the wrapped typed optional and checking that the fields included + the `fields` map + the implicit ones (contract address) gives the full configuration makes sense

packages/node-sdk/paima-db/src/sql/extensions.sql
```diff
@@ -10,18 +10,21 @@ SELECT * FROM chain_data_extensions
...
+    chain_data_extensions (
+        CDE_ID,
+        CDE_TYPE,
+        CDE_NAME,
+        CDE_HASH,
+        START_BLOCKHEIGHT,
+        SCHEDULED_PREFIX
+    )
+SELECT
+    COALESCE(:cde_id, MAX(chain_data_extensions.cde_id) + 1),
```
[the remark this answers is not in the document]
ecioppettini: I think there could actually be an issue with this, although I haven't tested this yet. Currently when loading the extensions on startup the engine will first load the ones from the file, and use the order to populate the `cdeId` field. Afterwards we load the dynamic extensions from the table. But if an extension gets added to `extensions.yml` I think the `cdeId` will clash with the one from the first dynamic primitive.

I'm not really sure how to handle this from the engine's point of view though, since it's more or less assumed that the `extensions.yml` shouldn't change.

I guess a way of making the Tarochi flow work could be to remove the dynamic primitives from the table, and adding them to `extensions.yml` before the new primitive.

packages/engine/paima-sm/src/types.ts
SebastienGllmt: On the docs side it would probably help to have an annotate image to explain how the settings are for different categories (dynamic triggering settings, field settings, wrapped primitive settings)

packages/node-sdk/paima-db/src/sql/extensions.sql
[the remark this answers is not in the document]
SebastienGllmt: The Tarochi flow is hacky and shouldn't be a long-term solution. Games need to update occasionally and that may require occasionally adding new primitive so ideally we would have better support for that. Having to do stuff manually for Tarochi is okay since we know how things work, but obviously this wouldn't really be an option for external teams

packages/engine/paima-sm/src/types.ts
ecioppettini: I moved the configs to a subfield already, it's now called `targetConfig`. I also renamed `fields` to `dynamicFields`, although I'm not sure if it's clear enough. It would look more like this now:

```yaml
extensions:
    - name: "dynamic nft"
...
    targetConfig: 
      scheduledPrefix: "nft"
      burnScheduledPrefix: "burn-nft"
      type: "erc721"
    dynamicFields:
      contractAddress: value
```

For some reason I can't use `Type.Partial` on the `ChainDataExtensionErc721Config` config though, I get a weird error: `Type instantiation is excessively deep and possibly infinite.`

As for this

> Why we hardcoded this to only ERC721 (it feels like everything should work with other standards as-is)

I wrote it with the idea of it being generic, I just reduced the scope to simplify testing and reviewing. But I can try extending this to the other ERC's. Testing every case would be super annoying though.

One issue is that I don't think it's possible to write the type to validate this correctly, but I can write a custom validation rule to check that `dynamicFields` + `config` provides a valid config.

packages/node-sdk/paima-db/src/sql/extensions.sql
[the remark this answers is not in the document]
ecioppettini: Yeah, I agree with that. It's just that currently there is no way of adding a primitive, so it's hard to say what that would look like.

Besides what I said before, there is also the fact that if you were allowed to edit the config file and the engine did a re-sync only for that primitive in order to get the state up-to-date, if you later did a re-sync from scratch then the `cde_id`s would also be different, since at that point we don't even know how many dynamic primitives would be there. This may not matter if nothing relies on this though, but it breaks determinism I guess (ignoring the fact that currently that would break determinism anyway because the new primitive may trigger state transitions in the past).

I guess something we could do here to future proof this is to assign cde ids to the dynamic primitives in a range that won't ever be used by the static ones? (negative numbers even?).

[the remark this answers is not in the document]
SebastienGllmt: > I guess something we could do here to future proof this is to assign cde ids to the dynamic primitives in a range that won't ever be used by the static ones? (negative numbers even?).

I think this will only partially be a solution because if you add a new dynamic primitive as part of an update, you may run into the same issue again, right?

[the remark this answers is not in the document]
ecioppettini: > I think this will only partially be a solution because if you add a new dynamic primitive as part of an update, you may run into the same issue again, right?

I don't follow why. What I'm saying is that the primitives in `extensions.yml` get assigned ids according to the order in the list like we do right now, for example in the range `[0; 100)`

And extensions created as a result of a dynamic trigger receive for example cde ids in the range `[100; ...)`. Of course using 100 just as an example to have small numbers.

Then if you start with a dynamic primitive trigger in your config, and 1 primitive is added dynamically, the ids `0` and `100` would be used. 

If you now upgrade and add a new trigger, it would have id `2`, so there is no collision there. And generated primitives would still be sequential, so the next one would be `101`.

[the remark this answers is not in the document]
ecioppettini: I think with the current code there should be a solution for this that only involves getting the cde ids from the cde name instead of relying on the order (except in the first-time setup). And if the engine knows that a primitive is new, then it's just a matter of querying the db to find the proper id.

But I think the name doesn't have to unique? And it doesn't solve the issue of a re-sync from scratch (if it's even an issue). It would also allow re-ordering the file, which may not be desired.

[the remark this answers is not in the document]
SebastienGllmt: > I don't follow why. 

Imagine your you have a dynamic primitive A in your codebase.
- At block 1111, A creates dynamic primitive ID 100
- At block 2111, A creates dynamic primitive ID 101

Now, you want to modify your game to include a new dynamic primitive B monitoring an existing contract for a new functionality. It's possible that this new dynamic primitive B has something at block 1112, which means you now end up with

- At block 1111, A creates dynamic primitive ID 100
- At block 1112, B creates *new* dynamic primitive ID 101
- At block 2111, A creates dynamic primitive ID **102** (note: this is no longer the same ID as before)

packages/engine/paima-sm/src/types.ts
SebastienGllmt: Okay, if leaving it to just ERC721 for now simplifies the type validation JS code we can leave it at that and create a Github issue explaining the limitation and the general idea on how to fix it (and then link to that issue from the docs)

packages/node-sdk/paima-db/src/sql/extensions.sql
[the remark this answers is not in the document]
ecioppettini: I see, right, that's a good point. That said, I don't think we really care about this as long as the id's are consistent in the db. I did some searching and I couldn't find any place where the id's are exposed to the game. It's certainly a problem if a new primitive got assigned the id of an existing one, which is what would happen right now if we naively allowed adding a primitive to `extensions.yml`,  but if the same primitive has a different id it shouldn't really matter as long as there are no clashes I believe.

The only potential difference is that it may lead to a different order in the extension processing maybe? Although currently the dynamic primitives are not sorted by any field, I imagine the sorting may be that of the primary key. But not sure if anything relies on this order.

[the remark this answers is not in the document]
SebastienGllmt: I think one way we can solve this with a similar solution is to say there are at most 32-bits work of cde IDs in the extensions.yml. Then, we say dynamic primitives always start with 4 bits of the parent ID, following by 48 bits for the order in which it was added (48 because max safe ints in Javascript are 52 bits)

42bits = 4,398,046,511,103 which should be more than enough. This also guarantees there are no conflicts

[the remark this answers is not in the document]
ecioppettini: why 4 bits? doesn't that limit you to 16 extensions? The 0th and the 16th would have the same 4 bits, so they would get the same namespace. I think for this to make sense we would need to use 52 - 32 = 20 bits for the split. I feel like 2 ** 20 = 1,048,576 dynamic primitives per dynamic trigger it's already a lot anyway.

Btw, we could also extract the bits from the hash... But I guess at that point we could just stop using incrementing ids and just use the hash as the id.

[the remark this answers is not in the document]
SebastienGllmt: Oh yeah my bad. It would have to be something more like 16 bits

Instead of using an incrementing number, using the hash of a primitive's configuration as an ID and not allowing 2 primitives with the same hash also sounds like it makes sense

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -36,26 +36,26 @@ import type {
 
 /* Functions to retrieve CDE ID: */
 
-export async function getCdeIdByName(
+export async function getcdeNameByName(
```
SebastienGllmt, line 39: I don't think this function does anything anymore because it takes the name as an input and returns the name as the output

packages/engine/paima-funnel/src/funnels/emulated/utils.ts
```diff
@@ -54,8 +54,8 @@ export function emulateCde(
     .flat()
     .sort((a, b) => {
       const bhDiff = a.blockNumber - b.blockNumber;
-      const cdeIdDiff = a.cdeId - b.cdeId;
-      return bhDiff ? bhDiff : cdeIdDiff;
+      const cdeNameDiff = a.cdeName.localeCompare(b.cdeName);
```
SebastienGllmt, line 57: this is a breaking change we'll have to mention

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
```diff
@@ -80,25 +80,25 @@ export async function internalGetAllOwnedNfts(
 
 export async function internalGetOwnedNfts(
```
SebastienGllmt, line 81: We'll probably want some other utility functions added around dynamic primitives. For example, if I'm monitoring 100 NFT contracts that I added dynamically, how do I get owners of all of them?

I think probably we just need something like:
1. Some way to know the number of IDs that exists for a dynamic primitive
2. A utility function to create a dynamic primitive cdeName from a base name + an ID

packages/engine/paima-funnel/src/reading.ts
```diff
@@ -178,21 +180,29 @@ export async function fetchDynamicEvmPrimitives(
     for (const _ext of exts) {
       const ext: CdeDynamicEvmPrimitiveDatum = _ext as CdeDynamicEvmPrimitiveDatum;
 
+      const cdeName = `${ext.cdeName}-${
```
SebastienGllmt: We should use re-use a utility function we expose publicly to combine a base name with an ID

i.e.

```js
generateDynamicPrimitiveName(
    ext.cdeName,
    sharedData.extensions.filter(
          e =>
            e.cdeType !== ChainDataExtensionType.DynamicEvmPrimitive &&
            e.cdeName.startsWith(ext.cdeName)
        ).length
)
```

packages/engine/paima-funnel/src/funnels/emulated/utils.ts
ecioppettini: mention where though? are we keeping a changelog?

packages/node-sdk/paima-utils-backend/src/cde-access-internals.ts
ecioppettini: > how do I get owners of all of them?

Currently you can use the function I added to get the dynamic primitives by parent, and extract the address from the config. This requires parsing the config though, but we could expose a wrapper otherwise.

> A utility function to create a dynamic primitive cdeName from a base name + an ID

How would you use this though? I'm not sure in which case you would want to use the numeric id, instead of getting the list and storing the names in case you want direct access later.

packages/engine/paima-funnel/src/funnels/emulated/utils.ts
SebastienGllmt: yeah I write the changelog when I release new versions. This comment is mostly for me to remember

SebastienGllmt (approved):
LGTM to merge. Any other utility function we decide to add can be done as a follow-up PR

PR #363, opened 2024-05-15
Pass Primitive transaction hash to STF for possible game use

SpaceManiac:
Games consuming Primitives might want to be able to know, for example to link to an explorer the specific transaction in which a mint occurred. To do that the information has to be stored and passed along somehow.

* Primitive datum objects now include the transaction hash that they originated from.
* A new table `scheduled_data_tx_hash` is added to be joined with `scheduled_data`. This is done instead of a nullable column to avoid having to introduce migration infrastructure.
* A new field `txHash?: string` is added to `STFSubmittedData` and passed to the state transition function provided by the game.

The game can ignore this field if it doesn't need it, and it doesn't bloat the concise encoding. However, the game must make use of the information immediately, as it's deleted along with the scheduled data once it's processed. Future expansion might be storing the transaction hash that minted or burned a given ERC-721 or ERC-1155 token into the relevant database table.


packages/node-sdk/paima-db/docker-postgres.sh
```diff
@@ -7,4 +7,4 @@ if ! docker compose up --wait; then
   docker compose logs --no-log-prefix
   exit 1
 fi
-npx pgtyped "$@"
```
SebastienGllmt, line 10: Did you mean to commit this?

packages/node-sdk/paima-db/src/scheduled-constructors.ts
```diff
@@ -10,12 +10,24 @@ import type {
...
-// Create an SQL update which schedules a piece of data to be run through
-// the STF at a future block height.
-export function createScheduledData(inputData: string, blockHeight: number): SQLUpdate {
+/**
+ * Create an SQL update which schedules a piece of data to be run through
+ * the STF at a future block height.
+ *
+ * @param inputData The input to pass to the STF, generally in Paima Concise format.
+ * @param blockHeight The future block height at which to post the input to the STF.
+ * @param txHash Transaction hash of the CDE event that triggered this input, for possible later use.
```
SebastienGllmt, line 19: FYI "CDE" is the deprecated name and we're trying to use the term `primitive` wherever possible. The reason for this is that we found the term `CDE` was too confusing for people not familiar with Paima (not only do they not know the acronym, "chain data extensions" isn't much easier for people to grasp)

SebastienGllmt (approved):
approving, but we should also update the docs to mention this feature

PR #364, opened 2024-05-17
Feature/stf chain parallel reorder

acedward:
# Draft

Pending 
- [ ] Export types
- [ ] Fix PR history (working on the wrong base branch)
- [ ] Test Parallel rules
- [ ] Test Retrocompatibility
- [ ] Test Reordering

This PR adds three related features to the STF Execution
* Create and Execute a `Serial` or `Parallel` Execution Plan for the STF.
* Adds an optional game callback `submittedDataReordering` function to reorder Submitted Inputs.
* STF Includes `ChainData` to allow STF to get additional information about the current block. 

## STF Chaindata argument
Adds new argument to STF invocations. 
This call is retrocompatible and does not require changes from the game side. 

Game STF signature 
```
export default async function (
  inputData: STFSubmittedData,
  blockHeight: number,
  randomnessGenerator: Prando,
  dbConn: Pool,
  chainData: ChainData
): Promise<SQLUpdate[]> {
```

## Reorder Function
Game STF Router can optionally return the submittedDataReordering function.
This is retrocompatible, the router can submit only stateTransition (does not require changes)

Game STF Router:
```
const reorderFunction = (chainData: { submittedData: { inputData: string }[] }) => {
  console.log('Reordering Submitted Data');
  return chainData.submittedData;
};
// This function allows you to route between different State Transition Functions
// based on block height. In other words when a new update is pushed for your game
// that includes new logic, this router allows your game node to cleanly maintain
// backwards compatibility with the old history before the new update came into effect.
function gameStateTransitionRouter(_: number): {
  stateTransition: Function;
  submittedDataReordering: Function;
} {
  return {
    stateTransition: gameStateTransitionV1,
    submittedDataReordering: reorderFunction,
  };
}
```

## Parallel Execution
ENV variable `GAME_STF_EXECUTION_MODE` can be set to `parallel` or `serial`
Serial by default.

This allows to run STF functions that do not have dependencies in parallel.
Parallel Execution

```
----------------------------------------
| Execution Plan (parallel):
|   @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|     @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|       @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|         @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|   @a|i|300| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|     @a|i|300| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|       @a|i|300| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|         @a|i|300| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
----------------------------------------


----------------------------------------
| Execution Plan (serial):
|   @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|     @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|       @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|         @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|           @a|i|301| (0xdc3da0d7a6338039d79272653172bdd056ec0f7a)
|             @a|i|2| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|               @a|i|2| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|                 @a|i|2| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
|                   @a|i|2| (0xbcb878950cbcfc4a1d13fd481ffeec6c2f310ff1)
----------------------------------------
```


SebastienGllmt:
We're going to need to rebase this since the orderbook-dex PR has been merged,  and also we need to update the Paima Engine docs to mention these features

packages/engine/paima-sm/src/index.ts
```diff
@@ -383,88 +422,266 @@ async function processScheduledData(
...
+      parents: [],
+      executed: false,
+    };
+
+    if (!keys.length) {
+      node.parents = [treeRoot];
+      treeRoot.children.push(node);
+    } else {
+      const blocked = node.keys
+        .map(k => treeEndNodes.find(e => e.key === k))
```
SebastienGllmt, line 588: I don't think this is going to work. The point of the parallel notation is that all entries that use the same identifier all get run sequentially

It looks like what you have implemented instead is that the first entry with this identifier gets executed, and then every other input that uses that identifier gets executed in parallel. Also, your implementation is O(n^2) in the number of inputs to process which isn't ideal

I think probably it's better to use a library like https://www.npmjs.com/package/dependency-graph 

1. `addNode` for every `submittedData` where the node name is just the index
2. Keep a mapping `<key, largest index containing that key>`
3. For every submittedData, find all its keys, and `addDependency` to the latest value in the mapping of (2)
Now, when executing, do
1. `nextNodes = overallOrder(true)`
2. `await nextNodes`
3. `nextNodes.forEach(node => removeNode(node))`

This should have better performance and be easier to read than managing our own execution node system

packages/contracts/evm-contracts/contracts/orderbook/IOrderbookDex.sol
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: MIT
```
SebastienGllmt, line 1: You need to rebase since these files are already merged into master

packages/engine/paima-sm/src/index.ts
SebastienGllmt: Also note there is an issue with the current implementation of the STF where a single error thrown by an STF call drops all transactions in a block. Ideally this would be changed so that failures only drop the submittedData that caused errors (easy change: just move the `SAVEPOINT` before around for loop for the `sqlQueries`)

The reason I bring this up is because it's less clear how this savepoint system works when parallelism mode is turned on. Most likely it would have to be something like this:
1. Run all the STF in parallel to get the list of queries to executive (if any of the STF failed, just ignore it)
2. Execute all the SQL queries in parallel after all the STF are done
3. If any of the SQL queries fails, exclude all sql queries generated from that `submittedData` and try again

However it's not immediately clear to me know we would implement (3) since I'm not sure how we can know which SQL query specifically was the cause of the crash and how we can know which `submittedData` it came from (recall: if one query crashes, all other queries that happen after it will crash I think with an error saying the transaction is in a bad state)

PR #367, opened 2024-05-23
Fix improper CommonJS import from web3-utils

SpaceManiac:
When running a `jest` test that ultimately imports this, I get an error like:

> SyntaxError: The requested module 'web3-utils' does not provide an export named 'sha3'

This seems to be because Node 20 / `cjs-module-lexer` 1.3.1 chokes on the `module.exports = {...}` object of `web3-utils` and ends up not including `sha3` as a distinct import. The solution is to import the default export and access its `.sha3` field normally. Long-term it might be better to upgrade web3-utils.

Discovered while working on https://github.com/PaimaStudios/paima-game-templates/pull/72.

### DIFF
--- a/packages/paima-sdk/paima-concise/src/batcher.ts
+++ b/packages/paima-sdk/paima-concise/src/batcher.ts
@@ -1,5 +1,5 @@
 import type { AddressType, WalletAddress, UserSignature, InputDataString } from '@paima/utils';
-import { sha3 } from 'web3-utils';
+import web3utils from 'web3-utils';
 
 export const OUTER_BATCH_DIVIDER: string = '\x02';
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
I feel like I remember there was a reason we did this where not doing it like this broke something else

Our goal though is to remove everything related to the web3 library though. It's an absolute garbage library and the entire industry is moving away from it

SpaceManiac:
Since it seems like this has some risk of mysteriously breaking stuff and I only ran into the problem in Jest tests or when running `tsc` output directly, which is a corner case, I'll close this in favor of someday replacing `web3` entirely.

PR #368, opened 2024-05-24
Isolate STF failures to the single input that caused them

SebastienGllmt:
This should fix the issue I mentioned in Slack and #364 where a single STF error can cause the entire block transaction to get skipped.

I haven't tested this PR, but hopefully it works


acedward (approved):
👍

PR #376, opened 2024-05-27
add dynamic extension names to stf

ecioppettini:
# About

This PR adds a new `extensionName` field in the submitted data when the primitive is a dynamically generated one. For example:

```
{
  userId: 0,
  realAddress: '0x0',
  userAddress: '0x0',
  inputData: 'burnNft|0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc|3',
  inputNonce: '',
  suppliedValue: '0',
  scheduled: true,
  scheduledTxHash: '0xba205481caa9529b6846fd62eead0a36c56b5ad3254d6713f7b320e5424a7b72',
  extensionName: 'Dynamic erc721##0'
} parsing input data
```

Also adds a new utility function to get the actual config by using this extension name field.

### DIFF
--- a/packages/engine/paima-sm/src/cde-generic.ts
+++ b/packages/engine/paima-sm/src/cde-generic.ts
@@ -1,4 +1,4 @@
-import { ChainDataExtensionDatumType, ENV } from '@paima/utils';
+import { ENV } from '@paima/utils';
 import type {
   CdeGenericDatum,
   CdeMinaActionGenericDatum,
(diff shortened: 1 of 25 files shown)


packages/engine/paima-sm/src/cde-generic.ts
```diff
@@ -15,14 +15,23 @@ export default async function processDatum(
   const blockHeight = cdeDatum.blockNumber;
   const payload = cdeDatum.payload;
   const prefix = cdeDatum.scheduledPrefix;
+  const shouldIncludeName =
```
SebastienGllmt, line 18: what's the rationale for not including the name?

ecioppettini: just that it's not really needed for non-dynamic primitives, since you already known which one is it from the scheduled prefix. 

that said, the join needs to be done anyway since it's a different table, so it doesn't really save anything right now by not being there. I can remove this logic and always add the field if it's simpler that way.

SebastienGllmt: yeah I think it's simpler to just always have it. Adds no computational complexity and might be useful for explorers

SebastienGllmt (approved):

PR #381, opened 2024-06-06
Feature/recaptcha

acedward:
This PR enables Google Recaptcha V3 for Batcher input submissions.
Only validated captcha results are posted into the blockchain.

To enable validation:
1. Create a reCaptcha3 account and get the `site-key` and `secret-key`. (https://www.google.com/recaptcha)
  * Set `RECAPTCHA_V3_BACKEND` in the .env.<NETWORK> file with your `secret-key`.
  * Set `RECAPTCHA_V3_FRONTEND` in the .env.<NETWORK> file with your `site-key`.
2. Add the reCaptcha3 script `<script src="https://www.google.com/recaptcha/api.js?render=${reCAPTCHA_site_key}" />` in the frontend html or call injectReCaptchaToHTML()

### DIFF
--- a/packages/batcher/webserver/src/index.ts
+++ b/packages/batcher/webserver/src/index.ts
@@ -20,6 +20,7 @@ import { createMessageForBatcher } from '@paima/concise';
 import { AddressType, getWriteNamespace, wait } from '@paima/utils';
 import { hashBatchSubunit } from '@paima/concise';
 import { GlobalConfig } from '@paima/utils';
+import axios from 'axios';
 
 const port = ENV.BATCHER_PORT;
 
@@ -88,6 +89,7 @@ type SubmitUserInputRequest = {
   game_input: string;
   timestamp: string;
   user_signature: string;
+  captcha: string;
 };
 type SubmitUserInputResponse =
   | {
@@ -333,13 +335,15 @@ async function initializeServer(
           return;
         }
 
-        const [addressType, userAddress, gameInput, millisecondTimestamp, userSignature] = [
-          req.body.address_type || AddressType.UNKNOWN,
-          req.body.user_address || '',
-          req.body.game_input || '',
-          req.body.timestamp || '',
-          req.body.user_signature || '',
-        ];
+        const [addressType, userAddress, gameInput, millisecondTimestamp, userSignature, captcha] =
+          [
+            req.body.address_type || AddressType.UNKNOWN,
+            req.body.user_address || '',
+            req.body.game_input || '',
+            req.body.timestamp || '',
+            req.body.user_signature || '',
+            req.body.captcha || '',
+          ];
         const input: BatchedSubunit = {
           addressType,
           userAddress,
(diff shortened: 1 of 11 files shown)


packages/batcher/webserver/src/index.ts
```diff
@@ -451,6 +460,45 @@ async function initializeServer(
   );
 }
 
+//
+// This method executes Google reCaptcha3 validation for backend.
```
SebastienGllmt, line 464: Comment blocks (`/*`) are recommended for function documentation in JS

packages/batcher/webserver/src/index.ts
```diff
@@ -419,6 +424,10 @@ async function initializeServer(
             );
             setWebserverClosed();
           } catch (err) {
+            if ((err as Error).message === 'Recaptcha validation failed') {
```
SebastienGllmt, line 427: Probably it's better to define a custom error type and use `instanceof` instead of this

packages/batcher/webserver/src/index.ts
```diff
@@ -451,6 +460,45 @@ async function initializeServer(
   );
 }
 
+//
+// This method executes Google reCaptcha3 validation for backend.
+// If the validation fails, an error is thrown - otherwise, the method returns void.
+// To enable validation:
```
SebastienGllmt, line 466: This is the kind of stuff we should mention in the Paima docs. I don't really mind repeating it here, but it will be duplicate comments. I tend to prefer putting it in the docs and then linking to the docs from the docs (it does mean the link in the code breaks if the doc url changes, but usually this isn't that big of a deal imo since knowing it's somewhere in the docs is usually enough)

packages/batcher/webserver/src/index.ts
```diff
@@ -451,6 +460,45 @@ async function initializeServer(
...
+  const recaptchaResponse = await axios({
+    url: recapchaURL,
+    method: 'post',
+    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+    data: {
+      secret: ENV.RECAPTCHA_V3_BACKEND,
+      response: captcha,
+    },
+  });
+  /* reCaptcha API Response example
```
SebastienGllmt, line 489: Do you want to define a proper Typescript type for this (maybe there is already type definitions you can download from somewhere? Otherwise, you can just define the type yourself even if it's not perfect)

packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
```diff
@@ -65,6 +65,39 @@ export async function updateFee(): Promise<void> {
   }
 }
 
+//
+// reCaptcha to validate batcher submissions
+//
+// To enable validation:
```
SebastienGllmt, line 71: the fact this comment block is duplicated in this function makes me think even more this should be in the docs instead

packages/batcher/webserver/src/index.ts
acedward: I didn't find a public type definition for this

SebastienGllmt (approved):
LGTM other than the fact we need docs for this feature as well

acedward:
FYI https://github.com/PaimaStudios/paima-engine-docs/pull/45

PR #383, opened 2024-06-07
Add PaimaParser.Json() helper

SpaceManiac:
The [Mina Primitive docs][1] currently advise just copy-pasting this wherever you need it. It's also used for generic EVM primitives. Seems like something useful to put in the shared toolbox.

[1]: https://docs.paimastudios.com/home/react-to-events/primitive-catalogue/mina/generic/#concise-format

### DIFF
--- a/packages/paima-sdk/paima-concise/src/PaimaParser.ts
+++ b/packages/paima-sdk/paima-concise/src/PaimaParser.ts
@@ -201,7 +201,7 @@ export class PaimaParser {
     // Add standard - common expressions to parse * and |
     grammar += `
 asterisk  ::= "*"
-pipe ::= "|" 
+pipe ::= "|"
 at ::= "@"
 `;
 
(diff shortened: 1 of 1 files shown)


SebastienGllmt (approved):
LGTM. The reason this never got done originally is because using just `JSON` means you lose type checking and we were hoping we could move the parsing system over to some system where the types are generated from a custom-defined parser (instead of now where the types and the code are duplicated)

However, we've ended up just documenting `parsimmon` as an option in the paima docs and ended up using `@sinclair/typebox`  inside Paima Engine itself without recommending it elsewhere. However, even if we do move to these systems, a lot of people will probably not care and just want the option that requires the least amount of code so having a "just json" option will help

SebastienGllmt:
FYI the docs you would have to modify for this PR are here: https://docs.paimastudios.com/home/read-write-L2-state/base-format#defining-types

PR #385, opened 2024-06-11
add avail funnels

ecioppettini:
There are two funnels added here. The first one is an analog to the evm parallel funnel, and has to be configured as `avail-other`:

```yaml
Avail:
  type: avail-other
  lightClient: 'http://localhost:7007'
  rpc: ws://127.0.0.1:9944
  delay: 200
  confirmationDepth: 10
```

The other one is an analog to the block funnel, and it has the same configuration, except that it has the key of `avail-main`.

In both cases the data is added as part of the submittedData field. Example

```
{
  realAddress: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
  inputData: 'avail|test1',
  inputNonce: '0xdacc51013196e97e83236c92370b8b0b1fe43a2cc4c6b2ee17a768932def1a7f',
  suppliedValue: '0',
  scheduled: false,
  userAddress: '5grwvaef5zxb26fz9rcqpdws57cterhpnehxcpcnohgkutqy',
  userId: 1
}
```

## Notes

This means the submitted data has to be utf8 text (I think it's how the evm version works too).

`suppliedValue` is always '0' because I don't think there is a concept of sending more tokens in a data submission. I think probably that would require setting up an extra transaction, and then submitting an input with the tx info or something (because tracking all the transfers may not work well with the data index?).

There is no presync phase for these funnels.

The delay on the base funnel version is something that we don't have in the block funnel. I don't know if it's needed here, but in any case it's there (defaults to no delay and no confirmation depth).

A light client is used for access to the app's data, but the rpc is used to get the slots for the blocks (from the headers). This is used to compute timestamps without relying on the steady block production. (Also see https://github.com/availproject/avail-light/issues/574)

# TODO
- [x] Refactor to re-use the code of the parallel evm funnel? The functions are mostly already abstracted, it's just a matter of replacing some.
- [x] Optimization? Only get headers for blocks with data (and latest one).
- [x] Batcher support? I think that's for a different PR. For now inputs have to be directly submitted. We may have already part of the cryto since it's polkadot.
- [x] There is a TODO in switchWallet, not sure what needs to be done there yet.
- [ ] Test with the emulated funnel just in case.
- [x] There are some warnings printed on init:


packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -147,6 +180,13 @@ const minaConfigDefaults = {
   delay: 30 * 40,
 };
 
+const availConfigDefaults = {
+  funnelBlockGroupSize: 100,
+  confirmationDepth: 10,
+  delay: 10 * 20,
```
SebastienGllmt, line 186: is there a rationale for this? Theses seem high

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -10,6 +10,8 @@ export enum ConfigNetworkType {
   EVM_OTHER = 'evm-other',
   CARDANO = 'cardano',
   MINA = 'mina',
+  AVAIL_MAIN = 'avail-main',
+  AVAIL_OTHER = 'avail-other',
```
SebastienGllmt, line 14: I think right now in the docs we don't talk about the avail-main option, right?

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
```diff
@@ -151,10 +151,16 @@ export class CarpFunnel extends BaseFunnel implements ChainFunnel {
       lastTimestamp = cachedState.timestamp;
     } else {
       lastTimestamp = await timeout(
-        this.sharedData.web3.eth.getBlock(blockHeight - 1),
+        this.sharedData.mainNetworkApi.getBlock(blockHeight - 1),
         DEFAULT_FUNNEL_TIMEOUT
       );
 
+      if (!lastTimestamp) {
+        // FIXME: this funnel shouldn't actually be doing this, it should be
```
SebastienGllmt, line 159: ?

packages/engine/paima-funnel/src/funnels/avail/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+  lc: string,
+  from: number,
+  to: number
+): Promise<{ blockNumber: number; submittedData: SubmittedData[] }[]> {
+  const data = [] as { blockNumber: number; submittedData: SubmittedData[] }[];
+
+  for (let curr = from; curr <= to; curr++) {
+    const responseRaw = await fetch(`${lc}/v2/blocks/${curr}/data?fields=data,extrinsic`);
+
+    // TODO: handle better the status code ( not documented in the api though ).
```
SebastienGllmt, line 95: It would be good to create an issue on the Avail repo/docs to mention documenting this if you think it would help, then linking to that issue from this code

packages/engine/paima-funnel/src/funnels/avail/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+  // FIXME: why is the conversion needed?
+  const header = (await api.rpc.chain.getHeader(hash)) as unknown as Header;
+
+  const slot = getSlotFromHeader(header, api);
+  return slotToTimestamp(slot, api);
+}
+
+export function slotToTimestamp(slot: number, api: ApiPromise): number {
+  // this is how it's computed by the pallet
+  // https://paritytech.github.io/polkadot-sdk/master/src/pallet_babe/lib.rs.html#533
```
SebastienGllmt, line 42: note: here and below, these are not permalinks, so this will stop working if they ever modify these files. Probably it's better to use Github permalinks instead

packages/engine/paima-funnel/src/funnels/avail/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export async function getLatestBlockNumber(api: ApiPromise): Promise<number> {
+  let highHash = await api.rpc.chain.getFinalizedHead();
+  let high = (await api.rpc.chain.getHeader(highHash)).number.toNumber();
+  return high;
+}
+
+export async function getTimestampForBlockAt(api: ApiPromise, mid: number): Promise<number> {
+  const hash = await api.rpc.chain.getBlockHash(mid);
+  // FIXME: why is the conversion needed?
```
SebastienGllmt, line 33: TODO?

packages/engine/paima-funnel/src/funnels/avail/parallelFunnel.ts
```diff
@@ -0,0 +1,445 @@
...
+      this.chainName,
+      InternalEventType.AvailLastBlock
+    );
+
+    return chainData;
+  }
+
+  private async waitForBlocksToBeProduced(latestKnownBlock: number): Promise<void> {
+    while ((await this.updateLatestBlock()) === latestKnownBlock) {
+      await delay(500);
```
SebastienGllmt, line 237: rationale for this number?

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: talk in which way? it is documented as a possible value, and it's also documented that it's exclusive with `evm-main`, but maybe it's purpose it's not clear enough if you don't deduce what it does through the analogue with the `evm-main`.

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: not really (except for the 20, that's the slot duration), I don't really know what's a sensible default for this. From what I've seen block finalization is usually 3 or 4 blocks behind, but I don't know if that will also be valid for mainnet.

SebastienGllmt: yeah finally should be relatively fast even on mainnet. You can ask in the Slack if you want to confirm with the team

packages/engine/paima-funnel/src/funnels/avail/utils.ts
ecioppettini: Yeah, but I'm not really sure how to fix this. I get this error otherwise:

>  src/funnels/avail/utils.ts(37,34): error TS2345: Argument of type 'import("/home/enzo/Work/paima-engine/node_modules/avail-js-sdk/node_modules/@polkadot/types/interfaces/runtime/types").Header' is not assignable to parameter of type 'import("/home/enzo/Work/paima-engine/node_modules/@polkadot/types/interfaces/runtime/types").Header'.
>  Property '#private' is missing in type 'import("/home/enzo/Work/paima-engine/node_modules/avail-js-sdk/node_modules/@polkadot/types/interfaces/runtime/types").Header' but required in type 'import("/home/enzo/Work/paima-engine/node_modules/@polkadot/types/interfaces/runtime/types").Header'.

This probably happens because the types don't come from the same package (even if I think these are the same, since `avail-js-sdk` uses the polkadot package). But there is only one version of `@polkadot/types` in this package, so I'm not sure what's the conflict.

packages/engine/paima-funnel/src/funnels/carp/funnel.ts
ecioppettini: I didn't want to change this here since it's unrelated, and it probably doesn't make a huge difference anyway. Basically the parallel evm funnel and the mina funnel store a checkpoint in the db to know where to resume pulling data after a restart (we need to get everything with a timestamp > than the latest synchronized block).

The carp funnel instead is instead getting (again) the previous block through the base funnel's rpc. It's not much of a deal because it's just done once, but I prefer the other option.

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: we could also just not provide any defaults?

SebastienGllmt: I think providing defaults makes sense (there is only one Avail chain and I don't imagine there being more than 1), so we just need to provide sensible defaults and a rationale for them

packages/engine/paima-funnel/src/funnels/avail/utils.ts
ecioppettini: Makes sense. There are basically two issues here:

1. I didn't see any docs for the status code. Not much of an issue because I can read the code, but it makes me a bit uncertain about how it could change.
2. If a block doesn't have _any_ submitted data (from you or from anyone) it returns a 400 (bad request) instead of a 404. This means we need to treat bad request the same as 404 (which is returned if you ask for a block in the future I think, but we should never do that).

packages/paima-sdk/paima-utils/src/config/loading.ts
ecioppettini: I was thinking about that and I started to realize that the `confirmationDepth` setting doesn't really make that much sense here. Avail (and Polkadot) have deterministic finality, and I think the light client doesn't give you blocks that are not finalized anyway. We still need the delay setting to account for this, for which we need a default, but I think I'll just remove this offset

SebastienGllmt: Okay, maybe good to mention that in the docs

packages/engine/paima-funnel/src/funnels/avail/utils.ts
SebastienGllmt: Okay, so do you want to create an issue and link to it from our code?

ecioppettini: done

SebastienGllmt (approved):

PR #386, opened 2024-06-13
Fee system for the Orderbook DEX

SebastienGllmt:


packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -120,8 +176,18 @@ contract OrderbookDex is IOrderbookDex, ERC1155Holder, ReentrancyGuard {
...
+                order.assetId,
+                orderId,
+                order.seller,
+                msg.sender,
+                assetsToBuy,
+                order.pricePerAsset,
+                makerFee,
+                takerFee
+            );
             if (remainingEth == 0) {
```
SebastienGllmt, line 191: What about if the value is something really close to zero, but not exactly zero because of floating point precision?

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -155,18 +222,48 @@ contract OrderbookDex is IOrderbookDex, ERC1155Holder, ReentrancyGuard {
...
+                order.assetId,
+                orderId,
+                order.seller,
+                msg.sender,
+                assetsToBuy,
+                order.pricePerAsset,
+                makerFee,
+                takerFee
+            );
             if (remainingAsset == 0) {
```
SebastienGllmt, line 267: Same comment

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -91,27 +123,51 @@ contract OrderbookDex is IOrderbookDex, ERC1155Holder, ReentrancyGuard {
...
             uint256 orderId = orderIds.unsafeMemoryAccess(i);
-            Order storage order = orders[orderId];
+            Order storage order = orders[asset][orderId];
             if (order.assetAmount == 0) {
                 continue;
             }
-            uint256 assetsToBuy = remainingEth / order.pricePerAsset;
+
+            uint256 purchaseCost = (remainingEth * basisPoints) / (order.takerFee + basisPoints);
+            if (purchaseCost * (order.takerFee + basisPoints) != (remainingEth * basisPoints)) {
```
SebastienGllmt, line 141: we should probably add a comment explaining what this line is for

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
matejos: It will continue iterating through orders, and then at the end if the minimum received asset (which has slippage subtraction applied) is satisfied, fill is successful and any remaining eth is returned to buyer

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
matejos: Same answer, although asset does not have decimals so the situation is less probable. Also floating point precision shouldn't make any problems if frontend properly operates with BigInt

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
matejos: Added

packages/contracts/evm-contracts/contracts/orderbook/OrderbookDex.sol
```diff
@@ -52,6 +56,14 @@ contract OrderbookDex is IOrderbookDex, ERC1155Holder, Ownable, ReentrancyGuard
         orderCreationFee = _orderCreationFee;
     }
 
+    /// @inheritdoc IOrderbookDex
+    function claim() external {
+        uint256 amount = balances[msg.sender];
+        delete balances[msg.sender];
+        emit BalanceClaimed(msg.sender, amount);
+        payable(msg.sender).sendValue(amount);
+    }
```
SebastienGllmt, line 65: should we only allow people to claim their own balance? It might be useful if another wallet can force-claim on behalf of somebody else

If you don't think this is useful in practice for any situation, I'm fine keeping it as-is

matejos: I can imagine some situations where that might be useful, but since we just agreed on making the contract upgradeable, let's leave the feature for future upgrades, if the need for it is there.

matejos (approved):

PR #389, opened 2024-06-26
Add mina-contracts package

SpaceManiac:
Adds `@paima/mina-delegation`:

- delegate.ts includes the `DelegationOrder`, `DelegationOrderProgram`, and `DelegationOrderProof` types that can be used as a common basis for Ethereum wallets to sign that a Mina address is generally authorized to act on their behalf, for use in ZK programs and contracts.
- The message under signature is a total of 74 bytes:
  1. The Ethereum standard prefix `\x19Ethereum Signed Message:\n`
  2. The Ethereum length-prefix `48`
  3. The prefix used by other Paima Engine delegations `DELEGATE-WALLET:`
  4. The Mina public key packed into 32 bytes
- Precaching the o1js circuit as part of the build is excluded in part due to its complexity, but mainly because the cached circuit amounts to 700 MB, which is unreasonably large for an NPM package.


acedward:
We have moved to Midnight for al ZK operations

Issue #373, opened 2024-05-27
Emit event standard for games


SebastienGllmt:
To avoid having to create custom indexers for every behavior, smart contract chains typically have an *event* standard, so that dApps can emit events in a way that every indexer and tool understands

Examples:
- EVM: https://medium.com/coinmonks/solidity-events-explained-82dc9104bc62
- SVM: https://www.rareskills.io/post/solana-logs-transaction-history

We should implement the same functionality in Paima for the same reason

PR #390, opened 2024-07-01
Paima events

acedward:
# New packages
`@paima/events`

This PR introduces an event system and also sets up the events to use `typebox` so that it is both strongly typed and so that we can later export them as json-schema for documentation purposes and so that it works well with the json-schema we'll use for the log system

- [x] doesn't solve the ENV issue
- [x] disable external users from sending messages to the engine/batcher (I believe this can be done with [this](https://github.com/moscajs/aedes/blob/main/docs/Aedes.md#handler-authorizepublish-client-packet-callback))
- [x] add more status options to the batcher event
- [ ] don't only send block events when a  block is non-empty (non-empty should maybe just be a filter on blocks?)
- [ ] fix the issue where block events may be skipped if paima engine shuts down (see `Handling shutdowns in event listeners` in https://github.com/PaimaStudios/paima-engine/issues/373)
- [x] Disable MQTT reserved keywords (`/`, `+`, `#`) in event names
- [x] test this PR

### Define an event

```typescript
const QuestCompletionEvent = genEvent({
  name: 'QuestCompletion',
  fields: [
    {
      name: 'questId',
      type: Type.Integer(),
      indexed: true,
    },
    {
      name: 'playerId',
      type: Type.Integer(),
    },
  ],
} as const);
```

### Listen to Messages

Listen to an event

```typescript
PaimaEventManager.Instance.subscribe(
  {
    topic: QuestCompletionEvent,
    filter: { questId: undefined }, // all quests
  },
  event => {
    console.log(`Quest ${event.questId} cleared by ${event.playerId}`);
  }
);
```

### Send Messages

Sent an event

```typescript
PaimaEventListener.Instance.sendMessage(
  QuestCompletionEvent,
  {
    questId: 5,
    playerId: 10
  }
);
```

`@paima/broker`

### Start MQTT Server
```typescript
new PaimaBroker(ENV).getServer('Paima-Engine');
```
This creates a WebSocket MQTT server  

# Tarochi Example (with Batcher)
https://github.com/user-attachments/assets/882e767b-6d77-4242-8b42-aa1003b40158

# Card-Game (Direct TX)
https://github.com/user-attachments/assets/4e0f6c59-b9d1-43bc-b558-c1855382f670

# Overview

image


packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -264,10 +275,29 @@ async function runSync(
...
+            const [e] = await emulatedSelectLatestPrior.run(
+              {
+                emulated_block_height: chainData.blockNumber,
+              },
+              gameStateMachine.getReadWriteDbConn()
+            );
+            emulated = block_chain;
+            block_chain = e.deployment_chain_block_height;
+          }
+          MQTTPublisher.sendMessage({ block: block_chain, emulated }, MQTTSystemEvents.STF);
```
SebastienGllmt, line 300: This event looks like it can be triggered multiple times for the same block. For the format of these topics, I think using the `Separated topics` section in https://github.com/PaimaStudios/paima-engine/issues/373 makes more sense

packages/engine/paima-runtime/src/runtime-loops.ts
```diff
@@ -422,3 +449,17 @@ async function blockPostProcess(
 
   return true;
 }
+
+const startMQTTBroker = async (): Promise<void> => {
+  if (ENV.MQTT_BROKER) {
+    // Load Server
+    MQTTBroker.getServer();
+    // TODO is this necessary?
+    await wait(1000);
```
SebastienGllmt, line 458: todo?

packages/mqtt/src/mqtt-broker.ts
SebastienGllmt, line 1: This file also looks like it's duplicated

packages/mqtt/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```
+npm run test
+```
+
+Lint:
+
+```
+npm run lint
+```
```
SebastienGllmt, line 27: Probably will have to update this file to just point to the docs once we have the MQTT setup specified in the Paima docs

tsconfig.base.json
```diff
@@ -27,6 +27,7 @@
     "noImplicitOverride": true,
     "skipLibCheck": true,
     "paths": {
+      // "@paima/mqtt/*": ["./packages/mqtt/*"],
```
SebastienGllmt, line 30: This won't work.

Packages in Paima need to be at depth 2 from the root folder (ex: `/packages/.../mqtt`) otherwise it won't work

packages/paima-sdk/paima-utils/src/mqtt/mqtt-events.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+export enum MQTTSystemEventsNames {
+  STF_GLOBAL = 'STF_GLOBAL',
+  BATCHER_HASH_GLOBAL = 'BATCHER_HASH_GLOBAL',
+  BATCHER_HASH_$ADDRESS = 'BATCHER_HASH_$ADDRESS',
+}
+
+export function newMQTTSystemEvent(
```
SebastienGllmt, line 25: What's the point of this function? It's not actually used afaik

packages/paima-sdk/paima-utils/src/mqtt/mqtt-client.ts
```diff
@@ -0,0 +1,111 @@
...
+      client.subscribe(event.path.fullPath, (err: any) => {
+        if (err) console.log(`MQTT[${event.broker}] ERROR`, err);
+      });
+    });
+
+    MQTTClient.subscriptions.push(event);
+  }
+
+  /* Helper to wait when hash is ready */
+  public static awaitBlock = async (hash: string, maxTimeSec = 20): Promise<number> => {
```
SebastienGllmt, line 95: It feels a bit weird that this function is inside the `MQTTClient`. Feels like we don't quite have separation of concerns

packages/batcher/batcher-transaction-poster/src/index.ts
```diff
@@ -95,16 +96,18 @@ class BatchedTransactionPoster {
       value: '0x' + Number(this.fee).toString(16),
       gasLimit: estimateGasLimit(msg.length),
     });
-    const receipt = (await transaction.extra.wait())!;
+    // TODO ONLY ACTIVATE IN ASYNC MODE!!!1!
```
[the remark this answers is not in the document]
acedward: This is something I'm still looking into. Waiting for confirmation adds a couple of seconds.

packages/batcher/batcher-transaction-poster/src/mqtt/mqtt-publisher.ts
```diff
@@ -0,0 +1,55 @@
+// import { ENV } from '@paima/utils';
+import mqtt from 'mqtt';
+
+const wait = async (n: number): Promise<void> =>
+  await new Promise(resolve => setTimeout(resolve, n));
+
+export enum MQTTSystemEvents {
+  BATCHER_HASH = '/sys/batch_hash',
```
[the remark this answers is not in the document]
acedward: https://docs.aws.amazon.com/whitepapers/latest/designing-mqtt-topics-aws-iot-core/mqtt-design-best-practices.html#:~:text=Since%20MQTT%20topics%20are%20case,when%20creating%20each%20topic%20level.

for paima events
/system/engine/stf
/system/batcher/hash/{hash}
...

for game events
/game/whatever

tsconfig.base.json
acedward: Thanks!

packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
```diff
@@ -104,16 +105,18 @@ export const postConciseData = async (
...
         PaimaMiddlewareErrorCode.ERROR_POSTING_TO_CHAIN,
         response.errorMessage,
         response.errorCode
       );
     }
-    const blockHeight = response.result;
-
+    if (typeof response.result !== 'number') {
+      throw new Error('Blockheight expected');
+    }
```
acedward, line 118: revert

packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
```diff
@@ -166,7 +200,7 @@ export async function postConciselyEncodedData(
         provider.getAddress(),
         gameInput
       );
-      return await submitToBatcher(subunit, captcha);
+      return await submitToBatcher(subunit, captcha, async);
```
acedward, line 203: not used, revert if not used

packages/paima-sdk/paima-events/src/system-events.ts
```diff
@@ -0,0 +1,106 @@
...
+      // Default system behaviors
+      switch (mqttEvent.name) {
+        //  eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+        case PaimaEventSystemName.STF_GLOBAL:
+          PaimaEventSystemParser.lastSTFBlock = data.block as number;
+          break;
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+        case PaimaEventSystemName.BATCHER_HASH_$ADDRESS:
+          PaimaEventSystemParser.hashes[data.hash as string] = data.blockHeight as number;
+          break;
```
acedward, line 37: This is not needed

packages/paima-sdk/paima-events/src/system-events.ts
```diff
@@ -0,0 +1,106 @@
...
+import { PaimaEvent } from './events.js';
+
+/*
+ * This class applies side-effects to system effects and then call the user defined callback
+ */
+export class PaimaEventSystemParser {
+  /* Shared storage for last blocked STF processed */
+  public static lastSTFBlock = 0;
+  /* Map of known hashes */
+  public static hashes: Record<string, number> = {};
```
acedward, line 13: remove

packages/batcher/batcher-transaction-poster/src/index.ts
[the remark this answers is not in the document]
acedward: I've set up an ENV variable for this.
There looks like no great solution, but it can take 3-4 seconds to get the confirmations.
When using the paima-events you want to know the input-hash quickly, and expect to be posted eventually.

packages/batcher/runtime/tsconfig.json
```diff
@@ -14,5 +14,6 @@
     { "path": "../../paima-sdk/paima-utils/tsconfig.build.json" },
     { "path": "../../node-sdk/paima-utils-backend" },
     { "path": "../../paima-sdk/paima-mw-core/tsconfig.json" },
+    // { "path": "../../node-sdk/paima-broker/tsconfig.build.json" },
```
SebastienGllmt, line 17: Is this meant to be commented out?

packages/batcher/utils/src/config.ts
```diff
@@ -127,7 +127,31 @@ export class ENV {
...
-  private static isTrue(value: string | undefined): boolean {
+  // MQTT BROKER
+  static get MQTT_BROKER(): boolean {
+    return ENV.isTrue(process.env.MQTT_BROKER, true);
+  }
+  static get MQTT_ENGINE_BROKER_PORT(): number {
+    return parseInt(process.env.MQTT_BROKER_PORT || '8883', 10);
+  }
+  static get MQTT_BATCHER_BROKER_PORT(): number {
+    return parseInt(process.env.MQTT_BROKER_PORT || '8884', 10);
```
SebastienGllmt, line 138: We have `8883` and `8884` hard-coded in many places despite them being specified here. I think it's because in a lot of places we return `undefined` as a possibility even though these `config` files are meant to avoid the undefined case

packages/batcher/utils/src/config.ts
```diff
@@ -127,7 +127,31 @@ export class ENV {
...
+  }
+  // MQTT CLIENT
+  static get MQTT_ENGINE_BROKER_URL(): string {
+    return process.env.MQTT_ENGINE_BROKER_URL || 'ws://127.0.0.1:8883';
```
SebastienGllmt, line 142: As an example of the above comment concretely, we shouldn't have `8883` here if the port is meant to be defined by the above ENV variable

packages/paima-sdk/paima-utils/src/config.ts
SebastienGllmt, line 1: Same comment here for isTrue and the ports as the other file

packages/batcher/utils/src/config.ts
acedward: This is the issue with the 'complex' interface.
This module can be called by paima-engine and the batcher.

They do not share a common interface so one alternative is:

```
import type { ENV as ENV1 } from '@paima/batcher-utils';
import type { ENV as ENV2 } from '@paima/paima-utils';

...
constructor(private env: typeof ENV1 | typeof ENV2) {}
...

env.I_CAN_USED_METHOD_DEFINED_IN_BOTH
```

But this made the Interface having to pass the global ENV (no very nice)

SebastienGllmt (approved):

Issue #336, opened 2024-03-27
Switch Chain Helper


acedward:
https://github.com/PaimaStudios/paima-engine/blob/b182760c08d2748b42c5bb3c447b720323bdcde0/packages/paima-sdk/paima-mw-core/src/wallets/evm/injected.ts#L26

SwitchChain gets the main game config by default, but adding/switching chains is required to interact with contracts (e.g., PRC-3 for minting/burning)

PR #391, opened 2024-07-02
avail batcher support

ecioppettini:
On top of #385 since otherwise it was not possible to test this.

There is one unrelated change here, which is that I added a way to have access to the configuration file in the middleware/frontend. The reason for this is that there is a `getBlockNumber`  utility that's used in the templates (at least in the chess) one, and I needed to know the network for that. Although I think the change it's probably good to have anyway.


packages/paima-sdk/paima-mw-core/src/wallets/evm/injected.ts
```diff
@@ -7,26 +7,40 @@ import { connectInjected } from '../wallet-modes.js';
...
-import { GlobalConfig } from '@paima/utils';
+import { ConfigNetworkType, GlobalConfig } from '@paima/utils';
 
 interface SwitchError {
   code: number;
 }
 
-async function switchChain(): Promise<boolean> {
-  const mainEvmConfig = await GlobalConfig.mainEvmConfig();
+async function switchChain(chainName: string | undefined): Promise<boolean> {
```
ecioppettini, line 16: I think this may also close #336 but not sure.

packages/paima-sdk/paima-mw-core/src/helpers/general.ts
```diff
@@ -1,7 +1,30 @@
+import { ConfigNetworkType, GlobalConfig } from '@paima/utils';
 import { getWeb3 } from '../state.js';
+import assertNever from 'assert-never';
 
 export async function getBlockNumber(): Promise<number> {
-  return await getWeb3().then(web3 => web3.eth.getBlockNumber());
+  const config = await GlobalConfig.mainConfig();
+  const networkType = config[1].type;
```
SebastienGllmt, line 7: I think `const [_, mainConfig]` is a bit more intuitive than using `config[1]` in multiple places

packages/paima-sdk/paima-mw-core/src/wallets/evm/injected.ts
```diff
@@ -7,26 +7,40 @@ import { connectInjected } from '../wallet-modes.js';
...
+  } else {
+    const evmConfig = await GlobalConfig.mainEvmConfig();
 
-  const [chainName, config] = mainEvmConfig;
+    if (!evmConfig) {
+      return true;
+    }
+
+    chainName = evmConfig[0];
+    config = evmConfig[1];
```
SebastienGllmt, line 37: nit: in my opinion, using [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) is usually cleaner and less error-prone than defining a `let` and remembering to set it in all branches

packages/batcher/batcher-transaction-poster/src/avail.ts
```diff
@@ -0,0 +1,35 @@
+import type { Pool } from 'pg';
+import { base64Encode } from '@polkadot/util-crypto';
+import BatchedTransactionPosterBase from './transactionPoster.js';
+
+class AvailBatchedTransactionPoster extends BatchedTransactionPosterBase {
```
SebastienGllmt, line 5: we will need to add information about how to use the Avail batcher to the Paima docs

packages/batcher/runtime/src/index.ts
```diff
@@ -166,31 +171,79 @@ async function main(): Promise<void> {
     'Game Batcher', // TODO: it doesn't matter now, but there is no way for the batcher to get the name of the game
     await getRemoteBackendVersion()
   );
-  const provider = await getWalletWeb3AndAddress(ENV.CHAIN_URI, privateKey);
+  let provider: EthersEvmProvider | AvailJsProvider;
+  let batchedTransactionPoster;
+  let getCurrentBlock;
+
+  if (!ENV.BATCHER_AVAIL_LIGHT_CLIENT) {
```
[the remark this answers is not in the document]
ecioppettini: I've been actually wondering if it wouldn't make more sense to just have one "binary" per network instead of having a single one where some environment variables work for one and others for another.

Although maybe the issue it's just that environment variables are really clunky for configuration.

In this case what would make sense to have, a variable that can be either `evm | avail | mina` ?

packages/batcher/batcher-transaction-poster/src/transactionPoster.ts
```diff
@@ -0,0 +1,193 @@
...
+    while (true) {
+      if (gameInputValidatorClosed && webserverClosed) {
+        process.exit(0);
+      }
+      await wait(1000);
+    }
+  };
+
+  protected postMessage = async (_msg: string): Promise<[number, string]> => {
+    throw new Error('unimplemented');
```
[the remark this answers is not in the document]
ecioppettini: Not really, but now that you mention it, this should be `abstract` since it's meant to be override.

packages/paima-sdk/paima-mw-core/src/helpers/general.ts
ecioppettini: I think that's better too, it's just that it also generates a 

```
6:10  warning  '_' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

and it's annoying to have to add an exception all the time.

ecioppettini: to be honest I regret making this thing returning an array instead of an object, and I was going to change it, but I didn't want to do it here since the PR already has a lot of changes.

packages/batcher/runtime/src/index.ts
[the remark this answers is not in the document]
SebastienGllmt: the executables are pretty large, so probably one exe per network isn't ideal filesize-wise

packages/paima-sdk/paima-mw-core/src/wallets/evm/injected.ts
SebastienGllmt: I think it does. We'll have to update the docs to mention this new feature

SebastienGllmt (approved):
LGTM modulo making sure we get the docs updated as well

Issue #388, opened 2024-06-16
Transaction hash standard


SebastienGllmt:
Right now, transactions in Paima don't have a globally unique hash. 

If we want a globally unique tx hash system, we have to consider the following cases for how the STF gets called:
1. a user submitting a transaction directly to an underlying chain
1. batcher transaction (that contains multiple underlying txs)
1. a primitive triggering on the underlying chain
1. scheduled data triggered by a timer

We do have some similar concepts though:
- Nonces for user inputs to an underlying chain are calculated using `[blockHeight] + [userAddress] + [game input]`
- Nonces from a batcher perspective are calculated using `[millisecondTimestamp] + [userAddress] + [game input]`
- `scheduledTxHash` which gives the underlying tx hash for primitives

However, this is incomplete as it doesn't handle all 4 cases.

Additionally, a better tx hash than these nonces could be something like `hash(caip2 prefix + underlying tx hash + idx)` where `idx` is because one tx in the underlying chain can trigger multiple STF calls (ex: trigger multiple primitives). This would handle all cases except for timers though. Note: `idx` here is not a global counter or a per-block counter, but rather a per-tx counter. This means that `idx` can be computer on the client side theoretically (if they know how many STF calls their transaction will trigger, or if they just query for `idx=0` from the node and check if any with higher IDs exist later)

For timers, we can do something like `blockNumber + contents`, but perhaps a different way to do it that is more tooling-friendly is to implement #387 first, then make that timers need a "percompile name" as an argument to set the address of these timers. This would allow us to do something like `address + blockNumber + idx` for the hash

Issue #387, opened 2024-06-16
Precompile standard


SebastienGllmt:
For things like events (#373) we need or EVM compatibility, we something needs addresses for things that are entirely in the rollup (not part of any underlying chain) and not a user address either. This is conceptually equivalent to precompiles you see in other chains (ex: Ethereum)

I think the best way to expose this is the following:

1. Games have a new folder called something like `precompile` (or maybe placed in the contracts folder)
2. The precompile folder has code that looks like this:
```js
import { generatePrecompile } from '@paima/...';

export const precompiles = {
    ...generaratePrecompile("foo"),
    ...generatePrecompile("bar")
} as const;
```
3. The JS code generates an object that looks like `{ Foo: "0x...", Bar: "0x..." }` where the hash value is something like `sha3("foo")`
4. In places in your code (ex: events) where you need to refer to these precompiles, you use `import { precompiles} from "contracts"; precompiles.Foo`
5. The precompiles need to be extracted and put in the `packaged` folder so that other tools can know the inverse map (the string that generated the address)

An example would be in Tarochi, when you buy something from the store, it could generate an event where the contract address is something like `generatePrecompile("general store")`

The API might need to be slightly different (ex: maybe use enums instead) to ensure that the types are resolved statically

PR #392, opened 2024-07-11
add txHash to the submittedData passed to the stf

ecioppettini:
Implements both #388 and #387, maybe we could split this in 2 PR's, but it was easier to test this way.

There are a few breaking changes here:

1. Added the network id to the mina config (for [this](https://github.com/ChainAgnostic/namespaces/blob/main/mina/caip2.md))
2. Added a `network` column to `scheduled_data_extension`.  This can be deduced from the `cdeName` column, so a migration shouldn't be too complicated if needed.
3. The signature of `createScheduledData`.


packages/node-sdk/paima-db/migrations/up.sql
```diff
@@ -17,7 +17,13 @@ CREATE TABLE scheduled_data_tx_hash (
 
 CREATE TABLE scheduled_data_extension (
   id INTEGER PRIMARY KEY REFERENCES scheduled_data(id) ON DELETE CASCADE,
-  cde_name TEXT NOT NULL
+  cde_name TEXT NOT NULL,
+  network TEXT NOT NULL
+);
+
+CREATE TABLE scheduled_data_precompile (
```
ecioppettini, line 24: I'm not really sure it makes sense to have this also as a new table instead of just adding a column tbh.

packages/node-sdk/paima-db/src/scheduled-constructors.ts
```diff
@@ -23,14 +23,15 @@ import type {
 export function createScheduledData(
   inputData: string,
   blockHeight: number,
-  cdeName?: string,
-  txHash?: string
+  source: { cdeName: string; txHash: string; network: string } | { precompile: string }
```
ecioppettini, line 26: does it really make sense to expose both use-cases in a public function? I think only the engine should use the cdeName part.

packages/paima-sdk/paima-precompiles/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@paima/precompiles",
```
SebastienGllmt, line 2: why did this need to be a standalone package? (maybe this is the right way to do it. Not sure the rationale)

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -310,3 +311,19 @@ function checkOrError<T extends TSchema>(name: string, structure: T, config: unk
...
+
+export function caip2PrefixFor(config: Static<typeof InstantiatedConfigsUnion>): string {
+  switch (config.type) {
+    case ConfigNetworkType.EVM:
+    case ConfigNetworkType.EVM_OTHER:
+      return `eip155:${config.chainId}`;
+    case ConfigNetworkType.MINA:
+      return `mina:${config.networkId}`;
+    case ConfigNetworkType.CARDANO:
+    // TODO: is this cip34 ?
```
SebastienGllmt, line 325: yes, cip34

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -310,3 +311,19 @@ function checkOrError<T extends TSchema>(name: string, structure: T, config: unk
   const decoded = Value.Decode(structure, config);
   return decoded;
 }
+
+const InstantiatedConfigsUnion = TaggedConfig(true);
+
+export function caip2PrefixFor(config: Static<typeof InstantiatedConfigsUnion>): string {
+  switch (config.type) {
```
SebastienGllmt, line 318: I think we need to add Avail as well once that PR is merged

packages/engine/paima-funnel/src/funnels/block/funnel.ts
```diff
@@ -179,6 +179,11 @@ export class BlockFunnel extends BaseFunnel implements ChainFunnel {
     }
   };
 
+  private caip2Prefix(): string {
+    // TODO: we may want to actually validate this against eth_chainId ?
```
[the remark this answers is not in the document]
ecioppettini: this is the evm block funnel, so it's not really possible for it to be anything else though

packages/engine/paima-sm/src/index.ts
```diff
@@ -224,20 +230,43 @@ const SM: GameStateMachineInitializer = {
 
         await processInternalEvents(latestChainData.internalEvents, dbTx);
 
+        // Used to disambiguate when two primitives have events in the same tx. This
+        // means the hash depends on the order we process the primitives (which is the
+        // order in the configuration).
+        // note that events triggered by the same tx may not be consecutive.
```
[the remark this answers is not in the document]
ecioppettini: I haven't really tested if it's the case since it's not that easy to setup, so I may be missing something, but as far as I understand from the current code we only group things by block at the funnel level. 

This means you can have 3 extensions that get triggered in the same block. And if two of them trigger in the same transaction, but the one that does not is in the middle of them (in the configuration order), the events here would also follow that order.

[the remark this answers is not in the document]
ecioppettini: I mostly added this comment because I initially had written the logic assuming that though, it's probably not really relevant.

packages/engine/paima-standalone/src/sm.ts
```diff
@@ -1,10 +1,17 @@
...
 import { poolConfig } from './utils/index.js';
 import { ENV } from '@paima/utils';
 
 export const gameSM = (): GameStateMachine => {
   const gameStateTransitionRouter = importGameStateTransitionRouter();
-  return PaimaSM.initialize(poolConfig, 4, gameStateTransitionRouter, ENV.START_BLOCKHEIGHT);
+  const precompiles = importPrecompiles();
+  return PaimaSM.initialize(
+    poolConfig,
+    4,
```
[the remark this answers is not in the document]
ecioppettini: it's been there since the beginning as far as I can tell? It's used for the prng's seed I think

packages/paima-sdk/paima-precompiles/package.json
ecioppettini: I had added it initially as a part of the utils package, but then the `precompiles.cjs` generated in the `packaged` directory in the game contained a lot of useless code from that package. I then changed it to this in order to reduce the size of that.

ecioppettini: There may be a way of eliminating that code though?

packages/engine/paima-sm/src/index.ts
```diff
@@ -44,9 +45,11 @@ import type {
   GameStateTransitionFunction,
   GameStateMachineInitializer,
   InternalEvent,
+  Precompiles,
 } from './types.js';
 import { ConfigNetworkType } from '@paima/utils';
 import assertNever from 'assert-never';
+import { sha3_256 } from 'js-sha3';
```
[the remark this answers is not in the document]
ecioppettini: I think so. I didn't test it myself, but the package has a browser demo https://emn178.github.io/online-tools/sha3_256.html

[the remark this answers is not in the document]
ecioppettini: It's also part of multiple dependencies already

packages/engine/paima-sm/src/index.ts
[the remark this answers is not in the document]
SebastienGllmt: Ooohh, I get it now

I read this as "may not" → cannot possibly happen (aka "shall not")
but you meant "may not" → there is a chance it won't happen

packages/engine/paima-funnel/src/funnels/block/funnel.ts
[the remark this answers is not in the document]
SebastienGllmt: ah, good point

packages/engine/paima-standalone/src/sm.ts
SebastienGllmt, line 12:
```suggestion
    4, // https://xkcd.com/221/
```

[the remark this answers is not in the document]
SebastienGllmt: ah, okay. I guess it's a reference to xkcd so I added a suggestion to at least document how this value was chosen

packages/paima-sdk/paima-precompiles/package.json
SebastienGllmt: Makes sense. Usually tree-shaking should get rid of unused stuff for production builds, but in my experience tree shaking is pretty unreliable to remove deadcode. We should at least document this rationale somewhere (possibly in the README for the project)

packages/node-sdk/paima-db/migrations/up.sql
SebastienGllmt: precompiles are not just scheduled data. They're for anything where you need to assign an address to something that isn't a real player (ex: buying from a shop could be considered sending funds to a precompile address)

ecioppettini: I'm not sure what that means in relation to this code. This is just to associate a particular scheduled input with a particular precompile in order to compute the hash. 

I'm not really storing the precompiles by themselves in the db, those are just imported from the packaged/precompiles.cjs and injected to the engine on launch.

ecioppettini: And the reason this is needed is because when you call `createScheduledData` you pass the precompile 'name' which is used as the address when the event is triggered.

packages/engine/paima-sm/src/index.ts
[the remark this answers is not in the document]
ecioppettini: >I read this as "may not" → cannot possibly happen (aka "shall not")

I actually didn't even know that was actual english. I could rephrase to `don't need` I guess.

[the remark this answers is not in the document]
ecioppettini: I don't think it's likely for that to happen in practice to be honest, but it's a bit counter-intuitive.

SebastienGllmt, line 236:
```suggestion
        // note: this means that events triggered by the same tx may, in some cases, not be consecutive.
```

SebastienGllmt (approved):

packages/engine/paima-sm/src/index.ts
```diff
@@ -358,25 +387,64 @@ async function processScheduledData(
...
+      let txHash;
+
+      const userAddress = data.precompile ? precompiles[data.precompile] : SCHEDULED_DATA_ADDRESS;
+
+      if (data.precompile && !precompiles[data.precompile]) {
+        doLog(`[paima-sm] Precompile for scheduled event not found ${data.precompile}. Skipping.`);
+        continue;
+      }
+
+      let caip2 = '';
```
SebastienGllmt, line 415: Is it okay that we leave this as empty string in one of the code paths? (being unsure about this is why I prefer IIFE over this kind of `let` approach)

ecioppettini: Well, it's empty if it's a timer... Since there is no specific network in that case. I guess we could otherwise make the property optional and just not include it in that case, but I'm not sure it makes much of a difference.

ecioppettini: I re-wrote it as an IIFE anyway though.

PR #393, opened 2024-07-15
Add middleware helpers for Mina

SpaceManiac:
Draft pending #389.

---

* Improve exported API of `mina-delegation` package to deal with usability issues.
  * Rename `DelegationOrder` to `DelegationCommand`.
* Add `MinaDelegationCache` to `@paima-sdk/providers`.
  * Given a `gameName`, this uses localStorage + a background worker to generate and cache a DelegationCommand proof.
  * Necessitated some build system fixups.
* Let some `IProvider.signMessage` implementations accept `Uint8Array` in addition to `string`.
  * Mina itself is excluded because Auro Wallet's API accepts strings only.
  * Therefore the `IProvider` interface itself only accepts `string` still.
  * It's still not advised to pass non-UTF-8 data, e.g. EVM `personal_sign` data is treated as text, so it won't look pretty.
  * Mainly included because of code improvements to how the strings get hex'd.

TODO:
* [ ] (?) Rename `mina-delgation` back to `mina-contracts` since caching is out & no longer demands that we split packages per program/contract.

Demo at https://github.com/PaimaStudios/paima-game-templates/pull/81.


acedward:
We have moved to Midnight for all ZK support

PR #394, opened 2024-07-16
Add @aiken-lang/aiken as a dev dependency

SpaceManiac:
Ran into this error while running `npm run build` during Mina work:

```
✖  nx run @paima/inverse-whirlpool:build
   > @paima/inverse-whirlpool@1.0.0 build
   > aiken build && aiken check -D
   
   sh: line 1: aiken: command not found
```

To solve, add the [@aiken-lang/aiken](https://www.npmjs.com/package/@aiken-lang/aiken) package as a devDependency of the Cardano contracts.

However, running `npm run build` after this does lead to a diff:
```
 packages/cardano-contracts/inverse-whirlpool/aiken.lock  |  4 ++--
 packages/cardano-contracts/inverse-whirlpool/plutus.json | 16 ++++++++--------
```
[aiken-build.txt](https://github.com/user-attachments/files/16241627/aiken-build.txt)

And I don't know enough about Aiken to say how much of a problem this is.


SebastienGllmt:
Oh I didn't know they published Aiken as an npm package as well. I just had it installed through aikenup instead

The modifications of aiken.lock and plutus.json are because I compiled them with `aiken v1.0.29-alpha+16fb02e` and you probably used some different alpha build

SpaceManiac:
My `npx aiken --version` also says `aiken v1.0.29-alpha+16fb02e`, although I do have the `x86_64-unknown-linux-gnu` version.
```
$ file node_modules/@aiken-lang/aiken/node_modules/.bin/aiken
node_modules/@aiken-lang/aiken/node_modules/.bin/aiken: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=98ab660d5ab1d7dfc25eff70a01c346f9f607d3c, for GNU/Linux 3.2.0, with debug_info, not stripped
```

SpaceManiac:
If `plutus.json` is a compile output, maybe it should be gitignored?

`aiken.lock` is a little odder - if it's like a package manager lockfile it should be included, but why then is it edited by every build?

SebastienGllmt:
Ideally there would be a way to keep both files, but I talked to the Aiken team and they recommend ignoring them both for now. Worst case, if projects depend on these files, we can use the `implicitDependencies` in nx to get these files generated.

SebastienGllmt (approved):

PR #396, opened 2024-07-16
Simplified events draft

SebastienGllmt:
This PR simplifies the event system, also sets up the events to use `typebox` so that it is both strongly typed and so that we can later export them as json-schema for documentation purposes and so that it works well with the json-schema we'll use for the log system

- [ ] doesn't solve the ENV issue
- [x] disable external users from sending messages to the engine/batcher (I believe this can be done with [this](https://github.com/moscajs/aedes/blob/main/docs/Aedes.md#handler-authorizepublish-client-packet-callback))
- [x] add more status options to the batcher event
- [ ] don't only send block events when a  block is non-empty (non-empty should maybe just be a filter on blocks?)
- [ ] fix the issue where block events may be skipped if paima engine shuts down (see `Handling shutdowns in event listeners` in https://github.com/PaimaStudios/paima-engine/issues/373)
- [x] Disable MQTT reserved keywords (`/`, `+`, `#`) in event names
- [x] test this PR

### Define an event

```typescript
const QuestCompletionEvent = genEvent({
  name: 'QuestCompletion',
  fields: [
    {
      name: 'questId',
      type: Type.Integer(),
      indexed: true,
    },
    {
      name: 'playerId',
      type: Type.Integer(),
    },
  ],
} as const);
```

### Listen to Messages

Listen to an event

```typescript
PaimaEventManager.Instance.subscribe(
  {
    topic: QuestCompletionEvent,
    filter: { questId: undefined }, // all quests
  },
  event => {
    console.log(`Quest ${event.questId} cleared by ${event.playerId}`);
  }
);
```

### Send Messages

Sent an event

```typescript
PaimaEventListener.Instance.sendMessage(
  QuestCompletionEvent,
  {
    questId: 5,
    playerId: 10
  }
);
```


packages/paima-sdk/paima-events/src/event-listener.ts
```diff
@@ -1,60 +1,118 @@
...
+export type CallbackAndMetadata<Event extends EventPathAndDef> = {
+  callback: (args: CallbackArgs<Event>) => void;
+  event: Event;
+};
 
-/* 
+/*
  * This class subscribes to specific topics, and stores the callbacks for event processing.
  */
 export class PaimaEventListener {
```
acedward, line 24: Now that this class works as Listener and Publisher, maybe rename just to PaimaEvents

PR #397, opened 2024-07-19
wrap all the funnels in config order across networks

ecioppettini:
This PR makes the events (funnel) order more consistent with what I documented in https://github.com/PaimaStudios/paima-engine-docs/pull/54

Currently the funnels are wrapped by the order of the configuration file, but there is a hardcoded order for the each funnel type. So if you set the funnels in the configuration like this:

```yaml

network1:
  type: mina
  ...

network2:
  type: other
  ...
``` 

The events from network2 will happen before network2. I think it's better to change things this way instead of trying to document that.

NOTE: we may want to merge #385 before this, since there will be a conflict and it may be easier to rebase this short branch.

### DIFF
--- a/packages/engine/paima-funnel/src/index.ts
+++ b/packages/engine/paima-funnel/src/index.ts
@@ -1,5 +1,15 @@
 import type { PoolClient } from 'pg';
-import { ENV, GlobalConfig, getPaimaL2Contract, initWeb3 } from '@paima/utils';
+import {
+  CardanoConfig,
+  ENV,
+  GlobalConfig,
+  MinaConfig,
+  OtherEvmConfig,
+  getPaimaL2Contract,
+  initWeb3,
+  validatePaimaL2ContractAddress,
+  AvailConfig,
+} from '@paima/utils';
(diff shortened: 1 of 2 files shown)


packages/engine/paima-funnel/src/index.ts
```diff
@@ -84,27 +88,53 @@ export class FunnelFactory implements IFunnelFactory {
...
+            this.sharedData,
+            dbTx,
+            ENV.START_BLOCKHEIGHT,
+            chainName,
+            config as CardanoConfig
+          );
+          break;
+        case ConfigNetworkType.EVM:
+          // the base funnel has to go first (even if it's not the first in the
+          // order) and it's already initialized
```
SebastienGllmt, line 131: Is this still true with the Avail branch being merged?

ecioppettini: I think so. I wrote base funnel instead of block funnel so that it's general. If the avail block funnel is used then it also goes first.

SebastienGllmt (approved):

PR #402, opened 2024-07-30
add BlockHeader type with timestamp info to the stf call

ecioppettini:
Adds a new block header type that gets passed on each stf call for that particular block, which has the block number and the timestamp.

### DIFF
--- a/packages/engine/paima-sm/src/index.ts
+++ b/packages/engine/paima-sm/src/index.ts
@@ -180,7 +180,7 @@ const SM: GameStateMachineInitializer = {
               caip2: '',
               txHash: '',
             },
-            blockHeight,
+            { blockHeight, timestamp: 0 },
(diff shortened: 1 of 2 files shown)


packages/engine/paima-sm/src/index.ts
```diff
@@ -180,7 +180,7 @@ const SM: GameStateMachineInitializer = {
               caip2: '',
               txHash: '',
             },
-            blockHeight,
+            { blockHeight, timestamp: 0 },
```
ecioppettini, line 183: we don't store the timestamps, so there is no easy way of getting the timestamp of the latest processed block. I'm not sure how much this matters though, since I'm not sure how the dry run is supposed to be used.

an easy alternative would be to put the current time maybe, but it's a bit weird.

SebastienGllmt (approved):

PR #405, opened 2024-08-02
initial work on complex-type events & asyncapi

SebastienGllmt:
I've added a new tool to generate `asyncapi` definitions based on the Paima events defined. Below is an example of what it looks like

image

I don't think `asyncapi` is as mature as swagger tooling, but it's at least something and gives us another tooling ecosystem we can tap into when needed since they do have quite a few tools already.

Additionally, before this PR, indexed events weren't supported on complex types (things that aren't strings/numbers). This PR does some of the initial work to allow this by making that complex data types are indexed on the hash of their value


packages/paima-sdk/paima-events/src/builtin-events.ts
```diff
@@ -56,9 +65,144 @@ const RollupBlock = genEvent({
...
-} as const satisfies Record<string, EventPathAndDef>;
+  const channels: Channels = {};
+  for (const [k, v] of Object.entries(events)) {
+    if (v.broker === PaimaEventBrokerNames.Batcher && info.batcherUri == null) {
+      continue;
+    }
+    const parameters: Parameters = {};
+    for (const param of v.path) {
+      if (typeof param === 'string') continue;
+      // TODO: properly handle enums
```
SebastienGllmt, line 137: TODO. This is notably helpful for the batcher status

PR #406, opened 2024-08-05
get avail slots/timestamps from light client instead of rpc

ecioppettini:
This issue https://github.com/availproject/avail-light/issues/574 was addressed already, so we can update the code here, so that we remove the need for the extra rpc requests to the node (since the light client already did those requests before).

We may want to wait for the changes to be in a release before merging this though, but I have already tested with the latest main.

There is also an issue with some types, but I'm not sure yet if there is a way to fix it properly.


acedward:
Avail has been reimplemented in Effectstream.
This might not be implemented in the new version, as funnel was used for inspiration.

PR #407, opened 2024-08-05
Add EVM-compatible JSON RPC wrapper to Paima

SebastienGllmt:
With this change, you can now query the current block number of a Paima rollup using EVM JSON RPC queries

```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' http://localhost:3333/rpc/evm
```

I've only implemented a few of the endpoints (some of them can't be implemented), but hopefully we can implement the few required to improve interoperability of Paima-based apps with existing tools

In the same PR, I also introduced a new typed `getPaimaNodeRestClient` so you can call `@paima/rest` endpoints with automatically generated type definitions instead of having to manually copy the types in the middleware everytime

# TODO

- [x] verify params actually match the types
- [x] `InternalRpcError` swallows the internal error which makes it not show up anywhere. It seems internal errors need a specific shape to work (see `RpcError`)
- [x] Fix the generated error types for `@paima/rest`

# Blockers

- [ ] add tx hash to the database (see #392 which calculates them, but doesn't store them anyway)
- [ ] create a block header standard and a block hash based off of it. Notably, we need to make sure it has the following properties:
    - [ ] it includes a timestamp (see #402 which adds it in-memory, but we need it from the database)
    - [ ] it includes a pointer to the previous block hash
- [ ] ensure that a timestamp field is present in block headers
- [ ] fetch logs for a transaction (see #409)
- [ ] a way to fetch the transaction nonce (low priority I think)
- [ ] A definition of "tx index" in a block. The natural definition for this is to sort by the database ID for values that have the same block, but this is tricky because there are 2 tables (`historical_game_inputs` and `scheduled_data`). The order of these IDs is also funnel-dependent which makes breaking changes frequent because of the way funnel fetch ordering works, but I don't think we can get around that. Right now we process the scheduled data in the STF before the historical data, so we can keep the same order
- [ ] Right now we delete scheduled data after it gets processed, but this breaks these historical queries


packages/engine/paima-runtime/src/evm-rpc.ts
```diff
@@ -0,0 +1,214 @@
...
+  MethodNotSupportedRpcError,
+  type EIP1193Parameters,
+  type PublicRpcSchema,
+} from 'viem';
+import { getPaimaNodeRestClient } from '@paima/mw-core';
+import { ENV } from '@paima/utils';
+
+export const evmRpcEngine = new JsonRpcEngine();
+
+// TODO: implement cache
```
SebastienGllmt, line 14: TODO

PR #409, opened 2024-08-06
add event logs to stf and db

ecioppettini:
Implements parts of #390 

There some missing parts still though:

- [x] Emitting the mqtt events.
- [x] Maybe expose some helper to use at the client side (although we may have this, I need to check)
- I'm not sure the current indexing make sense. We may want a composite index instead and always ask for all the indexed fields? But it's less versatile.
- The current rest interface has the issue that only 10 fields can be filtered. That said, I'm not sure how to this while leveraging pgtyped and checked arguments.
- [x] Also I imagine we want some block range size limitation.


packages/engine/paima-rest/src/controllers/BasicControllers.ts
```diff
@@ -397,3 +398,110 @@ export class TransactionContentController extends Controller {
     }
   }
 }
+
+type GetLogsResponse = Result<any[]>;
```
SebastienGllmt, line 402: Why `any[]` here? There are many known fields like
- block index
- tx index
- log index

packages/engine/paima-runtime/tsconfig.json
```diff
@@ -11,7 +11,6 @@
     { "path": "../../node-sdk/paima-db" },
     { "path": "../paima-rest/tsconfig.build.json" },
     { "path": "../paima-sm" },
-    { "path": "../../node-sdk/paima-broker/tsconfig.build.json" },
```
SebastienGllmt, line 14: we need this. Please don't remove it

packages/engine/paima-sm/src/index.ts
```diff
@@ -548,26 +643,66 @@ async function processUserInputs(
...
+
+          const eventDefinition = eventDefinitions[event.data.name].find(
+            eventDefinition => eventDefinition.topicHash === event.data.topic
+          );
+
+          if (!eventDefinition) {
+            throw new Error('Event definition not found');
+          }
+
+          eventsToEmit.push([eventDefinition, event.data.fields]);
```
SebastienGllmt, line 682: This whole block looks copy-pasted, right?

packages/engine/paima-standalone/src/utils/import.ts
```diff
@@ -67,3 +68,13 @@ const PRECOMPILES_FILENAME = 'packaged/precompiles.cjs';
...
 }
+
+const EVENTS_FILENAME = 'packaged/events.cjs';
+export type AppEventsImport = { events: ReturnType<typeof generateAppEvents> };
+
+/**
+ * Reads repackaged user's code placed next to the executable in `events.cjs` file
+ */
+export function importEvents(): AppEventsImport {
+  return importFile<AppEventsImport>(EVENTS_FILENAME);
```
SebastienGllmt, line 79: Probably for both this and the precompiles import, we should give a default value if the import fails (otherwise, this is a tedious breaking change for every existing Paima dApp to handle)

packages/node-sdk/paima-db/src/sql/events.sql
```diff
@@ -0,0 +1,51 @@
...
+  COALESCE(data->>:field0 = :value0, 1=1) AND
+  COALESCE(data->>:field1 = :value1, 1=1) AND
+  COALESCE(data->>:field2 = :value2, 1=1) AND
+  COALESCE(data->>:field3 = :value3, 1=1) AND
+  COALESCE(data->>:field4 = :value4, 1=1) AND
+  COALESCE(data->>:field5 = :value5, 1=1) AND
+  COALESCE(data->>:field6 = :value6, 1=1) AND
+  COALESCE(data->>:field7 = :value7, 1=1) AND
+  COALESCE(data->>:field8 = :value8, 1=1) AND
+  COALESCE(data->>:field9 = :value9, 1=1) AND
```
SebastienGllmt, line 12: So we only allow up to 10 fields in an event log max?

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+  T[number]['name']
+> & {
+  get: () => {
+    address: `0x${string}`;
+    data: { name: string; fields: KeypairToObj<T[number]['fields']>; topic: string };
+  }[];
+};
+
+// Typed stateful builder that handles overloaded topics. There is a performance
+// cost of using this, since it dinamically finds the matching overloaded topic
```
SebastienGllmt, line 96:
```suggestion
// cost of using this, since it dynamically finds the matching overloaded topic
```

packages/paima-sdk/paima-utils/src/config.ts
```diff
@@ -157,6 +157,10 @@ export class ENV {
     return process.env.MQTT_BATCHER_BROKER_URL || 'ws://127.0.0.1:' + ENV.MQTT_BATCHER_BROKER_PORT;
   }
 
+  static get GET_LOGS_MAX_BLOCK_RANGE(): number {
+    return parseFloat(process.env.GET_LOGS_MAX_BLOCK_RANGE || '5000');
```
SebastienGllmt, line 161: Hmm, do we need this? Seems a little arbitrary. If there's a rationale for this, we should definitely document it

packages/batcher/runtime/tsconfig.json
```diff
@@ -13,7 +13,6 @@
     { "path": "../../paima-sdk/paima-providers/tsconfig.build.json" },
     { "path": "../../paima-sdk/paima-utils/tsconfig.build.json" },
     { "path": "../../node-sdk/paima-utils-backend" },
-    { "path": "../../paima-sdk/paima-mw-core/tsconfig.json" },
-    { "path": "../../node-sdk/paima-broker/tsconfig.build.json" },
```
SebastienGllmt, line 17: don't remove this. We need it

packages/engine/paima-runtime/tsconfig.json
ecioppettini: is this warning a false positive then?

```
> @paima/root@0.0.0 lint:configs
> ts-node -P ./tsconfig.base.json ./tools/scripts/check-implicit-dependencies.ts

warn: package packages/engine/paima-runtime/tsconfig.json has unnecessary references
node-sdk/paima-broker
```

I removed it because of that and because it didn't seem to break anything, but I can undo the change. It would be nice to hide the warning if this is the case though.

packages/engine/paima-standalone/src/utils/import.ts
ecioppettini: Makes sense, I guess having no events should be a valid choice anyway (maybe not for precompiles). Currently when I don't care about it I just do `touch packaged/precompiles.cjs`, but it's not ideal, and wouldn't work for events I think.

packages/paima-sdk/paima-utils/src/config.ts
ecioppettini: the number is rather arbitrary, that said, I don't think it's generally a good idea to have unbounded queries? eth based nodes usually have a max range too. My concern is mostly DoS though. If this is covered by some other factor (a proxy?) I'll remove this. But isn't the engine's server usually queried by the clients directly?

packages/node-sdk/paima-db/src/sql/events.sql
ecioppettini: Not really a limitation of how many there can be, this is just how many (of the indexed ones) can you filter in a rest request.

We can lift this limitation though, it's just that I couldn't think of a way of making it work for an arbitrary number of fields with pgtyped. There are array expansions and object expansions, but I couldn't find a syntax that would let me write this kind of condition. 

My best idea is to build the query dynamically (with parameters) and use `pg.Client` query directly. And I can keep this query and cast the return value to that, since it should be the same type.

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+export function toSignature<T extends LogEvent<LogEventFields<TSchema>[]>>(event: T): string {
+  return event.name + '(' + event.fields.map(f => f.type.type).join(',') + ')';
+}
+
+export function toTopicHash<T extends LogEvent<LogEventFields<TSchema>[]>>(event: T): string {
+  return keccak_256(toSignature(event));
+}
+
+// this generates the type for the events import.
+export const generateAppEvents = <
```
ecioppettini, line 17: To contextualize this. The contents of  `events/index.ts`  would have to be something like:

```ts
import { genEvent } from '@paima/events';
import { generateAppEvents } from '@paima/events';

const StfEvent1 = genEvent({
  name: 'StfEvent',
  fields: [
    {
      indexed: true,
      name: 'tag',
      type: Type.String(),
    },
    {
      indexed: false,
      name: 'data',
      type: Type.String(),
    },
  ],
} as const);

const StfEvent2 = genEvent({
  name: 'StfEvent',
  fields: [
    {
      indexed: true,
      name: 'tag',
      type: Type.String(),
    },
    {
      indexed: false,
      name: 'data',
      type: Type.String(),
    },
    {
      indexed: false,
      name: 'extraData',
      type: Type.String(),
    },
  ],
});

const StfEvent3 = genEvent({
  name: 'StfEvent',
  fields: [
    {
      indexed: true,
      name: 'tag',
      type: Type.String(),
    },
    {
      indexed: false,
      name: 'data',
      type: Type.String(),
    },
    {
      indexed: false,
      name: 'extraData',
      type: Type.String(),
    },
    {
      indexed: true,
      name: 'block',
      type: Type.Integer(),
    },
  ],
});

const eventDefinitions = [StfEvent1, StfEvent2, StfEvent3] as const;

export const events = generateAppEvents(eventDefinitions);
```

(the contents don't really matter here, it's just part of my current test setup, but it's not finished).

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+      definition: event,
+      // we add this to avoid having to re-compute it all the time.
+      topicHash: toTopicHash(event),
+    });
+  }
+  return result as any;
+};
+
+// create payload for the stf from an object.
+export function createEventForStf<T extends ReturnType<typeof genEvent>>(
```
ecioppettini, line 49: This can be used like this in the stf function, for example:

```ts
const event = createEventForStf(user, StfEvent1, {
    tag: 'command1',
    data: 'test',
});

events.push(event);

return { stateTransitions: [], events  };
```

I'm open to suggestions for a better name for it though.

The main annoying part about this is function is that you need to name the overloaded events with some sort of unique suffix, and use the correct index.

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+//
+// queue.emit('A', data1);
+// queue.emit('B', data2);
+// queue.emit('B', data3);
+//
+// return { events: queue, stateTransitions: {...}};
+//
+// Where data2 and data3 may be different types (for the two overloaded
+// definitions), and the return type is that of the stf.
+export function eventQueueFactory<
```
ecioppettini, line 120: The api on this may be slightly better (so it's currently optional), but I think at least this may be a starting point

Currently I've been using this like this:

By previously initializing the object at the beginning of the stf function. Basically something like this:

```ts
export default async function (
  inputData: SubmittedChainData,
  blockHeader: BlockHeader,
  randomnessGenerator: Prando,
  dbConn: Pool
): Promise<{
  stateTransitions: SQLUpdate[];
  events: any;
}> {
  const events = eventQueueFactory.empty();

  const user = inputData.userAddress.toLowerCase();
  const parsed = parse(inputData.inputData);

  switch (parsed.input) {
    case 'command1':
      handleCommand1(events, user, parsed.tag, blockHeader.blockHeight);
      break;
    case 'input2':
      handleCommand2(events, user, blockHeader.blockHeight);
      break;
  }

  return { stateTransitions: [], events: events.get() };
}

function handleCommand1(
  eventQueue: ReturnType<typeof eventQueueFactory.empty>,
  user: string,
  tag: string,
  block: number
) {
  eventQueue.emit('StfEvent', user as any, { tag: 'command1', data: 'test', extraData: tag });

  eventQueue.emit('StfEvent', user as any, {
    tag: 'command2_1',
    data: 'test',
    extraData: 'extra',
    block: block,
  });
}
```

I think my preferred API would to pass an instance of this queue as an argument to the state transitition function, because I think having to return an object of arrays looks weird. The problem is that this currently can't be typed since the app events are loaded at runtime.

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+  T[number]['name']
+> & {
+  get: () => {
+    address: `0x${string}`;
+    data: { name: string; fields: KeypairToObj<T[number]['fields']>; topic: string };
+  }[];
+};
+
+// Typed stateful builder that handles overloaded topics. There is a performance
+// cost of using this, since it dynamically finds the matching overloaded topic
```
ecioppettini, line 96: I've been thinking that I included the topic because it was part of the spec/issue. But to be honest I'm not sure how much sense it makes. Maybe moving this logic to the engine would be better? And then this wouldn't be a concern at all. Because currently the user can submit an event with the wrong topic and there is no validation for it at the engine level (there is the validation that the topic exists for that name, but not that the actual fields match).

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
...
+        topic = topics[i].topic;
+        break;
+      }
+    }
+
+    if (!topic) {
+      // this shouldn't really happen because of the static typing, but it
+      // can't be inferred.
+      // it technically can also happen if there are extra fields, since
+      // typescript doesn't guarantee that.
```
ecioppettini, line 172: actually, I thought this was the case? But it doesn't seem to happen now that I tested it.

packages/engine/paima-sm/src/index.ts
```diff
@@ -591,6 +684,17 @@ async function processUserInputs(
   return latestChainData.submittedData.length;
 }
 
+async function sendEventsToBroker<Event extends EventPathAndDef>(
+  eventsToEmit: EventsToEmit<Event>
+) {
+  if (ENV.MQTT_BROKER) {
+    // we probably don't want to use Promise.all since it will change the order.
+    for (const [eventDefinition, fields] of eventsToEmit) {
+      await PaimaEventManager.Instance.sendMessage(eventDefinition, fields);
```
ecioppettini, line 693: I want to verify something just in case. Currently the emitted events look like this:

```
app/indexedField1/indexedField2 b'{"field1":"data","field2":"data"}'
``

This does seem to be consistent with the builtin events, but I'm not sure if not having the event name or some other identifier makes sense? Since I'm not sure how this would be parsed otherwise, unless a unique indexed field is used explicitly as a tag?

File: packages/engine/paima-runtime/tsconfig.json (hunk 2 above)
Comment (SebastienGllmt) [reply packages/engine/paima-runtime/tsconfig.json:14]: Yes, it's a false positive. It's an issue in `nx` itself afaik and the warning does have some implications about how nx handles things, so there is not much we can do other than ignore it

File: packages/paima-sdk/paima-utils/src/config.ts (hunk 7 above)
Comment (SebastienGllmt) [reply packages/paima-sdk/paima-utils/src/config.ts:161]: By default nodes are bounded, but archive nodes aren't bounded and those are usually what you get when you pay for an RPC provider. 

That is to say, I don't necessarily think have a default limit is bad (and you can have nodes that people pay to access if they want more historical data), but the reason Ethereum does it this was is because light nodes don't store this data at all (so you can't even get this data if you want to), which is different from our case where the node has all the logs anyways and is just refusing to serve them to you

i.e. I think our Ethereum equivalent would be to do something like say that if `STORE_HISTORICAL_GAME_INPUTS` is false, we only keep 5000 logs in the database or something like that

File: packages/paima-sdk/paima-events/src/app-events.ts (hunk 10 above)
Comment (SebastienGllmt) [reply packages/paima-sdk/paima-events/src/app-events.ts:49]: > The main annoying part about this is function is that you need to name the overloaded events with some sort of unique suffix, and use the correct index.

It's not clear to me why this would be required. This isn't required in Ethereum, so why wold it be required for us?

Comment (ecioppettini) [reply packages/paima-sdk/paima-events/src/app-events.ts:49]: I don't mean the name of the event itself, I mean the name of the javascript variable that has the definition. Or if you define them directly in an array, then you need the index in the array.

Although this is mostly because the topic is part of the input, and you need to know which overload it is (see this comment https://github.com/PaimaStudios/paima-engine/pull/409#discussion_r1719369411) to be able to provide it.

If we removed that requirement then I think we could have a simpler api. Although in that case we need to compute it from the data inside the engine, so that we know which type it is.

File: packages/paima-sdk/paima-utils/src/config.ts (hunk 7 above)
Comment (ecioppettini) [reply packages/paima-sdk/paima-utils/src/config.ts:161]: I see. I'm not sure why, but I was sure I had issues in the past with requesting logs from too many blocks in a single request (like for example, when setting the `funnelBlockGroupSize` too large in the evm funnel settings). And I also think I had problems with Besu at some point where I had to shrink the block size or it just didn't return anything, but maybe it was something else.

Anyway, maybe this is the wrong approach and instead we may want to paginate this endpoint (with a default max page size)? Although not sure who that would be adapted in the eth rpc wrapper then.

File: packages/engine/paima-standalone/src/utils/import.ts [hunk 15]
```diff
@@ -67,3 +68,17 @@ const PRECOMPILES_FILENAME = 'packaged/precompiles.cjs';
 export function importPrecompiles(): PreCompilesImport {
```
ecioppettini, line 68: TODO: add default for this too

packages/node-sdk/paima-db/src/event-indexing.ts
```diff
@@ -0,0 +1,27 @@
...
+  eventDescriptions: { topic: string; fieldName: string; indexed: boolean }[]
+): Promise<boolean> {
+  for (const event of eventDescriptions) {
+    const indexName = `index_${event.topic.slice(0, 20)}_${event.fieldName}`;
+
+    const checkQuery = `SELECT * FROM pg_indexes WHERE indexname = '${indexName}';`;
+
+    const result = await pool.query(checkQuery);
+
+    const createQuery = `CREATE INDEX ${indexName} ON event(topic, (data->>'${event.fieldName}'));`;
```
ecioppettini, line 14: TODO: add index for the topic

packages/node-sdk/paima-db/src/event-indexing.ts
```diff
@@ -0,0 +1,27 @@
+import type { PoolClient } from 'pg';
+
+export async function createIndexesForEvents(
+  pool: PoolClient,
+  eventDescriptions: { topic: string; fieldName: string; indexed: boolean }[]
+): Promise<boolean> {
+  for (const event of eventDescriptions) {
+    const indexName = `index_${event.topic.slice(0, 20)}_${event.fieldName}`;
```
ecioppettini, line 8: TODO: document this in the docs (one index per field)

packages/paima-sdk/paima-events/src/app-events.ts
```diff
@@ -0,0 +1,200 @@
+import type { LogEvent, LogEventFields } from './types.js';
+import type { genEvent } from './types.js';
+import { toPath, TopicPrefix } from './types.js';
+import type { TSchema, Static } from '@sinclair/typebox';
+import { Value } from '@sinclair/typebox/value';
+import { keccak_256 } from 'js-sha3';
```
ecioppettini, line 6: TODO: check the size of this in ethereum

packages/engine/paima-sm/src/index.ts
ecioppettini: TODO: add the topic/signatureHash as part of the prefix (but not for builtin events)

packages/engine/paima-rest/src/controllers/BasicControllers.ts
```diff
@@ -397,3 +398,151 @@ export class TransactionContentController extends Controller {
...
+    blockNumber: number;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: { [fieldName: string]: any };
+    tx: number;
+    idx: number;
+  }[]
+>;
+type GetLogsParams = {
+  fromBlock: number;
+  toBlock: number;
```
ecioppettini, line 415: TODO: make this optional

packages/node-sdk/paima-db/src/paima-tables.ts
```diff
@@ -753,6 +753,61 @@ const TABLE_DATA_CDE_DYNAMIC_PRIMITIVE_CONFIG: TableData = {
   creationQuery: QUERY_CREATE_TABLE_CDE_DYNAMIC_PRIMITIVE_CONFIG,
 };
 
+const QUERY_CREATE_TABLE_EVENT = `
+CREATE TABLE event (
+  id SERIAL PRIMARY KEY,
```
acedward, line 758: We shouldn't use `serial` fields here. The STF will eventually process inputs in parallel, making the order non-deterministic

packages/paima-sdk/paima-events/src/app-events.ts
ecioppettini: it's the same size (and not truncated), so if we want to make it shorter it would be different then

packages/node-sdk/paima-db/src/paima-tables.ts
ecioppettini: I can probably change the pk to block height + index inside the block instead, but I'm not sure it would make any difference. Can you explain what does it mean to process inputs in parallel? Because even then you could call the stf multiple times in parallel and then serialize the results.

packages/engine/paima-rest/src/controllers/BasicControllers.ts
```diff
@@ -397,3 +398,153 @@ export class TransactionContentController extends Controller {
...
+
+type GetLogsResponse = Result<
+  {
+    topic: string;
+    address: string;
+    blockNumber: number;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: { [fieldName: string]: any };
+    tx: number;
+    idx: number;
```
SebastienGllmt, line 410: It's probably not clear from the name that this is meant to represent the order in which this log was executed within the transaction

packages/engine/paima-sm/src/index.ts
```diff
@@ -462,31 +515,50 @@ async function processScheduledData(
...
+          eventsToEmit
+        );
       } catch (err) {
         // skip scheduled data where the STF fails
         doLog(`[paima-sm] Error on scheduled data STF call. Skipping`, err);
         continue;
       }
       if (sqlQueries.length !== 0) {
-        await tryOrRollback(DBConn, async () => {
+        const success = await tryOrRollback(DBConn, async () => {
```
SebastienGllmt, line 543: I don't think it's ever possible for `success` to be false here, given an error goes into the `catch` block

packages/engine/paima-sm/src/index.ts
```diff
@@ -548,26 +622,45 @@ async function processUserInputs(
...
+          eventsToEmit
+        );
       } catch (err) {
         // skip inputs where the STF fails
         doLog(`[paima-sm] Error on user input STF call. Skipping`, err);
         continue;
       }
       if (sqlQueries.length !== 0) {
-        await tryOrRollback(DBConn, async () => {
+        const success = await tryOrRollback(DBConn, async () => {
```
SebastienGllmt, line 650: same comment about success

packages/engine/paima-sm/src/index.ts
```diff
@@ -629,4 +733,46 @@ async function processInternalEvents(
...
+  for (let idx = 0; idx < events.length; idx++) {
+    const event = events[idx];
+    sqlQueries.push([
+      insertEvent,
+      {
+        topic: event.data.topic,
+        address: event.address,
+        data: event.data.fields,
+        tx: txIndexInBlock,
+        idx,
```
SebastienGllmt, line 761: again, not convinced of the name `idx` vs something more clear like `log_index`

packages/engine/paima-sm/src/index.ts
ecioppettini: Not false, but it should be `undefined` if it goes into the `catch` block, which is falsy. In that case it won't enter the if branch, which is the only thing that matters. Technically this could be a singleton object, but not sure there is a convenient one. Or we could change the `tryOrRollback` function to return `false` I guess, but then it wouldn't be generic on the function type.

packages/engine/paima-rest/src/controllers/BasicControllers.ts
ecioppettini: I renamed it to `logIndex`. Only caveat is that if we wrap this to implement `eth_getLogs`, we need to remember that this is the index of the log in the tx, and the other one is the index of the log in the block. I guess we may be able to compute the index of the log out of the tx index and the log index though?

ecioppettini: Or maybe we may want to change this field to be the log index in the block to be consistent with that? I just used the index in the tx because it's what was described in https://github.com/PaimaStudios/paima-engine/issues/373

SebastienGllmt: ah good catch. I feel it should probably by the index in the block instead of in the transaction then

packages/node-sdk/paima-db/src/paima-tables.ts
SebastienGllmt: I think having the ID be deterministic even in the case where transactions are executed in parallel is equivalent to the other open discussion on this PR: https://github.com/PaimaStudios/paima-engine/pull/409#discussion_r1728245329

SebastienGllmt (approved):
approving modulo the logIndex comment

PR #422, opened 2024-08-14
Batcher CDE

acedward:
https://github.com/user-attachments/assets/2d1f8d93-7208-43d6-8acf-e62be98e97a2

# Pending Paima Engine
- [x] API endpoint for game and batcher to check the current `balance` per `userAddress`/`batcherAddress`
- [ ] What to do with `cde_name` (or more than one BatcherPayment Contract)
- [ ] How do we distribute gas costs through multiple submissions in the same transaction? 
- [x] ENV flag to enable this behavior 
- [x] Emit (MQTT) Event when Batcher Payment event is triggered, to allow batcher to update the current balance.

# Pending Batcher
- [x] Check Balance and update/store in-memory current balance
- [ x call `getEstimatedFee` to temporarily subtract in-memory balance (to allow multiple TX) and update value at the mqtt/event
- [x] Reject input if there is not enough balance
- [x] ENV flag to enable this behavior


packages/engine/paima-sm/src/cde-evm-batcher-payment.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { batcherAddress, userAddress, value } = cdeDatum.payload;
+
+  const updateList: SQLUpdate[] = [];
+  console.log('processBatcherPayment', cdeDatum, inPresync);
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    updateList.push([
+      cdeBatcherPaymentUpdateBalance,
+      {
+        cde_name: 'generic-batcher-contract', // TODO Using constant name. Can there be more than one?
```
acedward, line 23: What to do with this.

packages/engine/paima-funnel/src/reading.ts
```diff
@@ -40,12 +42,21 @@ export async function getBaseChainDataMulti(
...
       block.timestamp,
       DBConn,
       caip2Prefix
     );
+
+    const batcherSubmittedData = submittedData.filter(s => s.fromBatcher);
+    if (batcherSubmittedData.length) {
+      const totalGasUsed = await getTotalGas(web3, blockEvents);
+      await splitGasCost(totalGasUsed, batcherSubmittedData, DBConn);
+    }
```
acedward, line 58: I'm not sure if this is the correct place for this, I didn't want to bubble up more data for now (gas used)

Or should we move this closer to the STF loop, and write in the main loop as well?

packages/engine/paima-sm/src/cde-evm-batcher-payment.ts
ecioppettini: I don't understand why is this a concern. There is a cdeName in the datum that can be used here. Which I imagine comes from the config. Even if there can be only one entry of this extension type for some reason, we should probably validate that instead.

packages/batcher/webserver/src/index.ts
```diff
@@ -405,49 +406,66 @@ async function initializeServer(
...
-                address_type: addressType,
-                user_address: userAddress,
-                game_input: gameInput,
-                millisecond_timestamp: millisecondTimestamp,
-                user_signature: userSignature,
-              },
-              pool
-            );
-            setWebserverClosed();
+            weiUsed = await BatcherPayment.estimateGasFeeInWei(JSON.stringify(payload));
```
SebastienGllmt, line 428: I think this code path is triggered even in non-EVM cases

packages/engine/paima-funnel/src/paima-l2-processing.ts
```diff
@@ -68,11 +73,11 @@ function decodeEventData(eventData: string): string {
 }
 
 export async function processDataUnit(
-  unit: NonTimerSubmittedData,
+  unit: NonTimerSubmittedData & { senderAddress?: string },
```
SebastienGllmt, line 76: If the goal of this is to track the contract address, that's now part of the `origin` field since #430

packages/engine/paima-sm/src/index.ts
```diff
@@ -384,6 +386,30 @@ async function processPaginatedCdeData(
...
+  let count = 0;
+
+  // Update batcher payments and fees.
+  for (const [query, params] of latestChainData.batcher?.sqlUpdates ?? []) {
+    count += 1;
+    await query.run(params, DBConn);
+  }
+
+  for (const e of latestChainData.batcher?.events ?? []) {
+    await PaimaEventManager.Instance.sendMessage(BuiltinEvents.BatcherPayment, {
```
SebastienGllmt, line 403: careful: it's possible for this database transaction to revert, in which case you've send events for a tx that never happened

packages/paima-sdk/paima-events/src/builtin-events.ts
```diff
@@ -68,9 +68,20 @@ const RollupBlock = genEvent({
   ],
 });
 
+const BatcherPayment = genEvent({
+  name: 'BatcherPayment',
+  fields: [
+    { name: 'batcherAddress', type: Type.String(), indexed: true },
+    { name: 'userAddress', type: Type.String(), indexed: true },
+    { name: 'operation', type: Type.Union([Type.Literal('payGas'), Type.Literal('addFunds')]) },
+    { name: 'wei', type: Type.String() },
```
SebastienGllmt, line 77: `wei` is EVM-specific, but the batcher is not just for EVM

There is maybe a way to make this more generic by adding the caip2 prefix, but you could also have different events for different stacks as well

packages/batcher/webserver/src/index.ts
acedward: Yes, it will fail as it has not been implemented.
Not sure what to do in this case.

https://github.com/PaimaStudios/paima-engine/pull/422/files/1ff0cf5cb1c7a7d0960632c8b353acda31e079c4#diff-a879925905151f317071cbb2b6633f93de28d49c8937bcd881c4b4efd933865bR64

packages/engine/paima-sm/src/index.ts
acedward: For the Batcher this should be expected, as gas fees should be charged as the TX was posted independent of the outcome (or we leave a loophole where a player could drain the batcher's funds by sending bad requests)

packages/paima-sdk/paima-events/src/builtin-events.ts
acedward: I updated the name to a more generic 'value'. As the batcher wallet has to be in the same network as the PaimaL2Contract, I think the caip2 might be redundant?

packages/engine/paima-funnel/src/paima-l2-processing.ts
acedward: Thanks for catching this, this is outdated 🙏  I've removed this address.

packages/engine/paima-sm/src/index.ts
SebastienGllmt: makes sense, but we should put a comment explaining this

acedward: added

packages/batcher/webserver/src/index.ts
acedward: Removed reference to `wei` units

PR #423, opened 2024-08-21
Add new Paima Block concept

SebastienGllmt:
This PR contains a somewhat controversial design decision on block hashes, so let me go over it

## What is this PR for?

Right now, Paima aggregates multiple stacks together, which is useful. The problem is that historically, we've had no way to refer to the aggregated state (only referring to the tx hash / block hash on the underlying stacks), or L2-specific concepts (like timers)

To solve this, we introduced transaction hashes in #392, and this PR is a continuation of that by introducing block hashes

## Decision 1) No Merklization of inputs

Typically, blocks contain a Merkle root of the inputs within a block, as it allows you to prove including of a transaction within a block  in logarithm space/time. However, how often are these Merkle trees actually leveraged, especially if it's for an app-specific game rollup? What is the performance overhead of these in JS? Are these even super relevant in the age of ZK?

Inputs in transaction *should* modify the block hash in some way, so let's explore three options:

### Merkle trees

the most common way to do this is via Merkle trees (combine tx hash in pairs as a tree until you have a single root)

I tried 2 different implementations:
1. A Merkle tree library optimized for EVM compatibility
2. A Merkle tree library that is unsafe for user-specified data (the data must come from deterministic sources) which enables some optimizations

based on my benchmark, if you have a game with 1000 txs/sec
1. Takes 45ms
2. Takes 3.5ms

Keep in mind, to have a 60fps game, you only have 16.6 ms to render each frame, so although the Merkle tree handling is a backend concern (so it doesn't effect frontend performance), conceptually it's equivalent to spending 20% of your budget just to generate the Merkle tree

The performance bottleneck, unfortunately, seems to be Javascript itself, as the Rust implementations of Merkle trees are 10~15x faster

### Iterated hashing

Iterated hashing works as follows:
1. Start a block with some initial seed `blockHash = hash(seed)`
2. For each txs in the block, do `blockHash = hash(blockHash || tx_hash)`

This approach halves the time down to 1.7ms
- Upside: faster
- Downside: to prove a tx is included in a block, you may have to iterate through every transaction from the starting hash (that is to say, O(n) performance instead of log(n) for Merkle trees)

### Single hash

Another alternative is to just hash the entire data together into one big blob: `hash(tx_hashes.join("|"))`
which brings the time down to 0.06ms
- Upside: faster
- Downside: requires all txs and an agreed-upon data format to calculate (trying to pass 1000 txs to Solidity to prove the inverse mapping is definitely not going to work as it's too much data)

### What is the most ZK-friendly?

There is a question of how much all of this provability matters. In an ideal future, you can just ZK prove things with the following way:
1. Start with a block hash that is agreed to be in the chain history
2. ZK-prove that the relevant data is included in a history that leads to that block hash

To accomplish this, all you need to do is run some ZK computation that checks the `previous block` pointer to go back in history until the point of interest, and then proves that some input exists in the block hash at this point in the past

Since zkVMs support private inputs, in the single-hash case you don't actually need to share the full data to prove a specific transaction is included in the proof, so it turns out this approach is quite ZK-efficient in this sense, compared to the Merkle tree approach which needs log(n) hash operations

Of course, note that this just gives you a ZK proof that an input exists, and says nothing about the resulting state from processing that input (so this is not super useful in practice anyway)

### Decision: what technique do we use?

Given that for games I feel we care about performance of creating blocks a lot more than the ability to Merkle-prove block content, I think just the single hash approach is probably the best option

## Decision 2) do we include failed txs in the block hash?

The problem with including failed transactions is that they are in theory not financially valuable. They can situationally be useful for debugging purposes, but they will never include the full set of failed txs some some txs may fail before even getting to the point of being given a transaction hash (although these cases are usually dev errors, as opposed to the txs we usually want to debug which are user-errors)

Despite this, I felt it's probably worth it to include failed transactions precisely because we didn't pick the Merkle tree approach, so including them is fairly cheap and so it's probably worth any debugging benefit we get from them


packages/engine/paima-sm/src/index.ts
```diff
@@ -51,10 +52,39 @@ import type {
...
+    prevBlockHash,
+    mainChainBlochHash: strip0x(chainData.blockHash),
+    blockHeight: chainData.blockNumber,
+    msTimestamp: chainData.timestamp * 1000,
+    successTxsHash: keccak_256(successfulTxs.join('|')),
+    failedTxsHash: keccak_256(failedTxs.join('|')),
+  };
+  return {
+    header,
+    hash: keccak_256(stableStringify(header)),
```
SebastienGllmt, line 80: out of all the decisions, using `stableStringify` here feels like the most questionable

packages/engine/paima-sm/src/index.ts
```diff
@@ -481,15 +554,17 @@ async function processScheduledData(
           }
         });
       }
+      resultingHashes.successTxHashes.push(txHash);
```
SebastienGllmt, line 557: this is wrong (here and below). We have to instead look at the return value of `tryOrRollback`

PR #425, opened 2024-08-23
Proposed changed to event export interface

SebastienGllmt:
The new code for the `events` package is as follows

```ts
import type { EventQueue } from '@paima/events';
import { registerEvents } from '@paima/events';
import { genEvent } from '@paima/events';

export const ClickEvent = genEvent({
  name: 'MyClickEvent',
  fields: [
    {
      indexed: true,
      name: 'user',
      type: Type.String(),
    },
    {
      name: 'time',
      type: Type.Number(),
    },
  ],
});

export const eventDefinitions = {
  ClickEvent,
} as const;

export const events = registerEvents(eventDefinitions);
export type Events = EventQueue<typeof eventDefinitions>;
```

You can then use it inside a subscribe call like this

```ts
import { events } from '@game/events';

await PaimaEventManager.Instance.subscribe(
  {
    topic: events.ClickEvent,
    filter: { user: undefined }, // all users
  },
  event => {
    console.log(`EVENT: ${event.time}`);
    callback(`Click time: ${event.time}`);
  }
);
```

### DIFF
--- a/packages/engine/paima-sm/src/index.ts
+++ b/packages/engine/paima-sm/src/index.ts
@@ -60,7 +60,7 @@ import type {
 import { ConfigNetworkType } from '@paima/utils';
 import assertNever from 'assert-never';
 import { keccak_256 } from 'js-sha3';
-import type { AppEvents, EventPathAndDef, generateAppEvents, ResolvedPath } from '@paima/events';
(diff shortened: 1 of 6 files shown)


ecioppettini (approved):
LGMT

PR #427, opened 2024-08-30
replace paima standalone packaged executable with a @paima/engine lib

ecioppettini:
This adds a @paima/engine package, that replaces the standalone binary. The packages exposes a `runPaimaEngine` function that can be used as an entry point.

This new package could be used like this:

## Example

```ts
import gameStateTransitionV1 from '@chess/state-transition';
import { runPaimaEngine } from '@paima/node-sdk/engine';
import RegisterRoutes from '@chess/api';
import openapi from '@chess/api/src/tsoa/swagger.json';

const events = {};

export enum PrecompileNames {
  Default = 'default',
}

export const precompiles = generatePrecompiles(PrecompileNames);

function gameStateTransitionRouter(blockHeight: number) {
    return gameStateTransitionV1;
}

(async (): Promise<void> => {
  await runPaimaEngine(gameStateTransitionRouter, precompiles, events, openapi, {
    default: RegisterRoutes,
  });
})();
```

This has multiple advantages over the current setup.

1. We don't need to ship node.
2. We can distribute through npm (or at least easier and with a small package).
3. This makes it easier to ensure that the game and the engine have to use the same version of node-sdk.
4. It probably produces a smaller final "executable" too, since in theory it should allow re-using some dependencies.
5. Easier to typecheck.

Also it's probably a bit easier to debug for developing.

### Bundling the game

I couldn't make chess work without creating a bundle with esbuild. I think this should be solvable, but there are some issues here and there. I didn't want to spend that much time on this since we may end up changing the template setup, but otherwise I think some changes may be needed there. In theory it should be possible to just run the game while using the node's normal module resolution logic.

When using esbuild (which is different than the current setup, because the game and the engine are basically bundled together), it's necessary to copy some of the assets to the bundle's directory. I added a script to help with this, but it's optional, since in theory it should be possible to use no bundler (or other bundler). Or also it can be avoided by using the `external` property of esbuild, for example. The script can be used like this:

```js
const { copyAssetsForBundle } = require("@paima/engine/scripts/copyAssets.cjs");

copyAssetsForBundle("packaged");
```

I think this is a mild annoyance compared to having the bundled _binary_ with everything, but I think the benefits are worth it. There is potentially another solution that involves providing an esbuild plugin, but I think it's too complicated to be worth it.

### node-sdk dependencies

I had to make public a lot of the needed dependencies for this to be easily used. There are workarounds for this, but it's not really that simple. This is slightly annoying because those packages can't really be used directly, but I think that's probably ok. On the bright side it makes debugging a bit easier sometimes, since you can just replace one of the modules (e.g. copy only @paima/funnels to a template while debugging a funnel).

### Configuration

If think potentially we could have the configuration be an argument to this function instead of reading the files. But I think there is already an issue about this, so I didn't want to tackle that here.

### Upgrades

I haven't really tested this, because I'd have to tweak the current setup for the local verdaccio registry, but I think potentially we could have an engine upgrade method that relies on npm aliases?

So for example, there could be multiple versions of `node-sdk` installed, and then a setup like this:

```ts
import { runPaimaEngine as runPaimaEngineV1 } from '@paima/node-sdk-v1/engine.js';
import { runPaimaEngine as runPaimaEngineV2 } from '@paima/node-sdk-v2/engine.js';

function gameStateTransitionRouter(blockHeight: number) {
  if (ENV.START_BLOCKHEIGHT <= blockHeight1) {
    return gameStateTransitionV1;
  }
  else return gameStateTransitionV2;
}

await runPaimaEngineV1(gameStateTransition, ..., { stopAt: blockHeight1 });
await gameMigrationLogicV2(); // if needed
await runPaimaEngineV2(gameStateTransition, ...,  { stopAt: null });
```

Of course this wouldn't really work right now since the engine doesn't really handle migrations, but that could be implemented. The other option would be to just split different versions in different packages at the game's side, create multiple bundles, and have an orchestrator that runs each one of them (potentially with different configs).


packages/node-sdk/paima-engine/scripts/copyAssets.cjs
```diff
@@ -0,0 +1,32 @@
+function copyAssetsForBundle(outDir) {
```
SebastienGllmt, line 1: `@paima/build-utils` is meant for stuff like this (so that the build tools can be a devDependency while still keeping the Paima node as a non-dev dependency in package.json)

packages/node-sdk/paima-funnel/src/funnels/BaseFunnel.ts
```diff
@@ -1,4 +1,8 @@
-import type { ChainFunnel, FunnelJson, ReadPresyncDataFrom } from '@paima/runtime';
+import type {
+  ChainFunnel,
+  FunnelJson,
+  ReadPresyncDataFrom,
+} from 'packages/node-sdk/paima-runtime/build/index.js';
```
SebastienGllmt, line 5: bad import style

SebastienGllmt (approved):

PR #428, opened 2024-08-31
Add funnel for Midnight

SpaceManiac:
Adds a funnel for [Midnight], using its [GraphQL API][gql] to monitor for and retrieve new blocks and their serialized contract states, which are then deserialized and converted to JSON by [@midnight-ntwrk/onchain-runtime][rt] and fed to scheduled as an input.

For example, the incrementing-counter example emits state updates that look like this, where the `2` is the actual number of increment calls:
```
{"tag":"array","content":[{"tag":"cell","content":{"value":[{"0":2}],"alignment":[{"tag":"bytes","length":8}]}}]}
```

Tested against fully-local Midnight chain using `standalone.yml` Docker Compose file from the Midnight counter [example].

Open questions:
* Is the JSON representation of the contract states stable enough?
* What's a "good default" for the confirmation depth?

TODO:
* [x] Use GraphQL WebSocket API to subscribe to new block events instead of polling on a timer
* [x] Convert Midnight CDEs to state machine inputs (`cdeTransitionFunction`)
* [x] Obey CDE configs instead of spamming each & every contract state update
* [x] Add DB table to remember last seen Midnight block height
* [x] Add confirmation depth config like Mina has, because we don't actually have instant finality
* [ ] Verify correctness of EVM<->Midnight block height interactions
* [ ] Test thoroughly
* [ ] Make logging less verbose

### Example

config.localhost.yml
```
MyMidnight:
  type: midnight
  indexer: http://localhost:32807/api/v1/graphql
  networkId: 0  # 0 for localhost, 1 for devnet
  confirmationDepth: 10
```
extensions.yml
```
extensions:
  - name: "my midnight extension"
    type: midnight-contract-state
    network: MyMidnight
    contractAddress: 010000c39bbf7ba21aa4c953e08412a8ff7ab1fab4a7cc11a00ad20eeb8050965ee838
    startBlockHeight: 0
    scheduledPrefix: test
```

[Midnight]: https://docs.midnight.network/
[gql]: https://docs.midnight.network/develop/reference/midnight-api/pubsub-indexer/#graphql-api
[rt]: https://www.npmjs.com/package/@midnight-ntwrk/onchain-runtime
[example]: https://docs.midnight.network/develop/tutorial/building/examples-repo


packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+  }
+`;
+
+class MidnightFunnel extends BaseFunnel implements ChainFunnel {
+  constructor(
+    sharedData: FunnelSharedData,
+    dbTx: PoolClient,
+    public chainName: string,
+    private readonly baseFunnel: ChainFunnel,
+    private config: MidnightConfig,
```
SebastienGllmt, line 166: to fit with #430
```suggestion
    public chainInfo: ChainInfo<MidnightConfig>,
    private readonly baseFunnel: ChainFunnel,
```

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+    this.nextBlockHeight = 0;
+  }
+}
+
+export async function wrapToMidnightFunnel(
+  chainFunnel: ChainFunnel,
+  sharedData: FunnelSharedData,
+  dbTx: PoolClient,
+  chainName: string,
+  config: MidnightConfig
```
SebastienGllmt, line 418:
```suggestion
  chainInfo: ChainInfo<MidnightConfig>,
```

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+        }
+      }
+    }
+
+    return result;
+  }
+
+  public override async readPresyncData(
+    args: ReadPresyncDataFrom
+  ): Promise<{ [network: number]: PresyncChainData[] | typeof FUNNEL_PRESYNC_FINISHED }> {
```
SebastienGllmt, line 284:
```suggestion
  ): Promise<{ [caip2: number]: PresyncChainData[] | typeof FUNNEL_PRESYNC_FINISHED }> {
```

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+        lastLogTime = now;
+        lastLogBlock = block.height + 1;
+      }
+
+      this.cache.nextBlockHeight = block.height + 1;
+
+      const extensionDatums = this.extensionsFromBlock(ENV.SM_START_BLOCKHEIGHT, block);
+      if (extensionDatums.length) {
+        result.push({
+          network: `midnight:${this.config.networkId}`,
```
SebastienGllmt, line 322:
```suggestion
          caip2: `midnight:${this.config.networkId}`,
```

You'll want to use the change you made to `caip2PrefixFor` instead of hardcoding it here

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+
+      block = await this.fetchBlock(this.cache.nextBlockHeight);
+    }
+
+    console.log(
+      `Midnight funnel ${this.config.networkId}: finished presync #${start}-${this.cache.nextBlockHeight - 1}`
+    );
+
+    const baseData = await baseDataPromise;
+    baseData[this.chainName] = result;
```
SebastienGllmt, line 336:
```suggestion
    baseData[caip2] = result;
```

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,431 @@
...
+    if (!startingBlock) {
+      throw new Error("Couldn't get main's network staring block timestamp");
+    }
+
+    const start = this.cache.nextBlockHeight;
+    let block = await this.fetchBlock(this.cache.nextBlockHeight);
+    if (!(block && midnightTimestampToSeconds(block.timestamp) < startingBlock.timestamp)) {
+      // We're caught up. Return that we've finished.
+      const baseData = await baseDataPromise;
+      baseData[this.chainName] = FUNNEL_PRESYNC_FINISHED;
```
SebastienGllmt, line 297:
```suggestion
      baseData[caip2] = FUNNEL_PRESYNC_FINISHED;
```

packages/engine/paima-sm/src/cde-midnight-contract-state.ts
```diff
@@ -0,0 +1,24 @@
...
+  cdeDatum: CdeMidnightContractStateDatum,
+  inPresync: boolean
+): Promise<SQLUpdate[]> {
+  const { scheduledPrefix, payload, blockNumber } = cdeDatum;
+  const updateList: SQLUpdate[] = [];
+
+  const scheduledBlockHeight = inPresync ? ENV.SM_START_BLOCKHEIGHT + 1 : blockNumber;
+  const scheduledInputData = [scheduledPrefix, payload].join('|');
+  updateList.push(
+    createScheduledData(scheduledInputData, scheduledBlockHeight, {
```
SebastienGllmt, line 16:
```suggestion
    createScheduledData(scheduledInputData, { blockHeight: scheduledBlockHeight }, {
```
among other changes to `createScheduledData` introduced in #430

packages/paima-sdk/paima-utils/src/config/loading.ts
```diff
@@ -131,6 +133,16 @@ export const AvailConfigSchema = Type.Intersect([AvailRequiredProperties, AvailO
...
 
+export const MidnightConfigSchema = Type.Object({
+  // Midnight network connection info.
+  networkId: Type.Number(),
+  indexer: Type.String(),
+  indexerWS: Type.Optional(Type.String()),
+  // node URL and proof server are not needed for our read-only use
+});
+
+export type MidnightConfig = Static<typeof MidnightConfigSchema>;
```
SebastienGllmt, line 144:
```suggestion
export const BaseMidnightConfigSchema = Type.Object({
  // Midnight network connection info.
  networkId: Type.Number(),
  indexer: Type.String(),
  indexerWS: Type.Optional(Type.String()),
  // node URL and proof server are not needed for our read-only use
});

export const MidnightConfigSchema = addType(BaseMidnightConfigSchema , ConfigNetworkType.MIDNIGHT);

export type MidnightConfig = Static<typeof MidnightConfigSchema>;
```
based on the changed required for #430

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,451 @@
...
+      }
+    }
+
+    // Insert the internal MidnightLastBlock event into each ChainData.
+    const composed = composeChainData(baseData, result);
+    for (const chainData of composed) {
+      (chainData.internalEvents ??= []).push({
+        type: InternalEventType.MidnightLastBlock,
+        caip2: this.caip2,
+        block: this.cache.nextBlockHeight - 1,
```
ecioppettini, line 378: does this work correctly if there is an exception in the middle of processing the returned (composed) data? Or if there is a crash. In the case where there are multiple blocks in baseData.

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,451 @@
...
+    const baseDataPromise = this.baseFunnel.readPresyncData(args);
+
+    const startingBlock = await this.sharedData.mainNetworkApi.getStartingBlock();
+    if (!startingBlock) {
+      throw new Error("Couldn't get main's network staring block timestamp");
+    }
+
+    const start = this.cache.nextBlockHeight;
+    let block = await this.fetchBlock(this.cache.nextBlockHeight);
+    if (!(block && midnightTimestampToSeconds(block.timestamp) < startingBlock.timestamp)) {
```
ecioppettini, line 287: just personal preference, but isn't this more clear as:

```ts
if (!block || midnightTimestampToSeconds(block.timestamp) >= startingBlock.timestamp) {
```

Although I'm not sure I understand why the block being null implies that it's finished.

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,451 @@
...
+          `Midnight funnel ${this.chainInfo.config.networkId}: presync #${lastLogBlock}-${block.height}`
+        );
+
+        lastLogTime = now;
+        lastLogBlock = block.height + 1;
+      }
+
+      this.cache.nextBlockHeight = block.height + 1;
+
+      const extensionDatums = this.extensionsFromBlock(ENV.SM_START_BLOCKHEIGHT, block);
```
ecioppettini, line 312: why does this have to be `SM_START_BLOCKHEIGHT` and not `START_BLOCKHEIGHT`? (not implying it's wrong)

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,451 @@
...
+    });
+    this.iter = this.wsClient.iterate<{ blocks: CachedBlock }>({
+      query: `subscription { blocks { ${cachedBlockQuery} } }`,
+    });
+    this.wsBlocks = [];
+  }
+
+  async load(dbTx: PoolClient): Promise<MidnightFunnelCacheEntry> {
+    const checkpoint = await getMidnightCheckpoint.run({ caip2: this.caip2 }, dbTx);
+    console.log(checkpoint);
```
ecioppettini, line 421: maybe want to remove this at some point?

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
```diff
@@ -0,0 +1,451 @@
...
+    if (checkpoint.length > 0) {
+      this.nextBlockHeight = Number(checkpoint[0].block_height) + 1;
```
ecioppettini, line 423: will this be 0 if this branch is not executed? does this mean the funnel will always traverse the entire chain? in the other funnels we have some code in place so that we only start from the first block we need (either we do a binary search on the evm chains, or we just compute the slot if we can).

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
SpaceManiac: `block` being null means we hit the head of the Midnight chain. Perhaps the presync should stall for new blocks just like regular sync in that case?

packages/engine/paima-funnel/src/funnels/midnight/funnel.ts
SpaceManiac: Good catch. I followed after the Mina and Carp examples but missed that they were actually supplying different heights for each output `ChainData`. Fixed.

SebastienGllmt (approved):
Approving because the integration Eddie will do will play the role of the testing

PR #431, opened 2024-09-09
fix usage of emulatedBlocksActiveOnBackend

ecioppettini:
I'm not entirely sure this is the right fix, but at the very least this seems to fix a bug where the frontend thinks the emulated blocks setting is always on (which breaks the confirmations code).

I'm not really sure what should happen if succes is false though.

### DIFF
--- a/packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
+++ b/packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
@@ -213,9 +213,11 @@ export async function postConciselyEncodedData(
 }
 
 async function getAdjustedHeight(deploymentChainBlockHeight: number): Promise<number> {
-  const activeOnBackend = await emulatedBlocksActiveOnBackend();
   const emulatedActive =
-    getEmulatedBlocksActive() ?? (activeOnBackend.success && activeOnBackend.result);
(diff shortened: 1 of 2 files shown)


packages/paima-sdk/paima-mw-core/src/helpers/posting.ts
```diff
@@ -213,7 +213,10 @@ export async function postConciselyEncodedData(
 }
 
 async function getAdjustedHeight(deploymentChainBlockHeight: number): Promise<number> {
-  const emulatedActive = getEmulatedBlocksActive() ?? (await emulatedBlocksActiveOnBackend());
+  const activeOnBackend = await emulatedBlocksActiveOnBackend();
+  const emulatedActive =
+    getEmulatedBlocksActive() ?? (activeOnBackend.success && activeOnBackend.result);
```
SebastienGllmt, line 218: this is a performance regression I think. The code on the left only calls the endpoint is `getEmulatedBlocksActive` is true, but the code on the right always calls the endpoint

ecioppettini: yeah, good point, I didn't consider the order. I'll fix that.

SebastienGllmt (approved):

Issue #415, opened 2024-08-10
sort `eth_getLogs` calls by `logIndex`


SebastienGllmt:
Right now, the Primitive processing flow for EVM looks like this:

1. Requests all primitives requires for a specific block range
2. Aggregate all the primitives returned from (1), and sort them by `blockNumber` (this is done in the function `groupCdeData`)

The problem is that blocks can contain many different logs inside them! We want to ensure that Paima Engine parses events from these logs in the same order they were emitted onchain, but `groupCdeData` doesn't do this at the moment. Fortunately, there is a field `logIndex` for exactly this purpose, and we just need to integrate it

PR #432, opened 2024-09-11
sort by logIndex for evm chains in groupCdeData

ecioppettini:
implements #415


SebastienGllmt:
This needs to be rebased

Issue #413, opened 2024-08-10
Scheduling events based on time instead of blocks


SebastienGllmt:
Currently, when calling `createScheduledData`, you need to pass a `blockHeight: number` of when to trigger the event. Of course, if you're using emulated blocks or a chain that has a deterministic block time, this works fine. However, not all networks have this and not all use-cases benefit from emulated blocks.

It is actually possible right now to do this even without emulated blocks. You can do this using the following trick:
1. Schedule an event for next block
2. Once the event from (1) triggers, check the timestamp in the header
3. If the timestamp is less than your target, just schedule an event again

Of course this works, but it's ugly to do this from a user perspective. It feels like something that is an engine-level feature

There are two ways we could handle this at the engine level:
1. **Reschedule every block**: just like mentioned above, we could reschedule the event every block on their behalf until the right time has come. All this requires is tracking in the database some new target timestamp
    1. **Advantage**: requires minimal code change
    2. **Disadvantage**: constantly recreating events like this messes with statistics and explorers for the game (ex: #411) because it's constantly spamming transactions (although at least they all get deleted)
1. **Create a new event type**: instead of using the scheduled data which requires a block, we should have some *new* event system that instead uses a timestamp (or make the block field optional in the existing system, along with a new timestamp field). Then, when it's time for an STF call, we check if there any events with a timestamp ready to use
    1. **Advantage**: we don't need to constantly re-create scheduled data. It can just be a single event that gets triggered when necessary
    1. **Disadvantage**: requires handling of this new block-less schedule data type

**Note**: it's not really possible to have timers *trigger* the STF without a block being made, as this functionally ends up being equivalent to the emulated block feature

## Additional consideration: timezones

If you want to schedule events at a specific time instead of a block, it's reasonable you may want to schedule for a specific timezone (ex: midnight PST). However, this has two unfortunate problems:

1. **Timezones change**: some areas have daylight savings so the time mapping changes during the year. Additionally, if some US states decide to abolish daylight savings, the mapping of Unix timestamp to midnight changes as well. In practice, as long as all nodes apply this change at the same time in the same way, it won't cause any issues.
2. **Tricky support**: to solve the problem mentioned in (1), we need a library to maps timezones in a way that is not only smart enough to handle things like daylight savings, but to *also* keep that library up-to-date to take into account any laws pass that affect timezones of importance

### Implementing timezones: `Date` api

[...]

PR #433, opened 2024-09-14
datetime based scheduled data/timers

ecioppettini:
see #413

### DIFF
--- a/packages/node-sdk/paima-db/src/scheduled-constructors.ts
+++ b/packages/node-sdk/paima-db/src/scheduled-constructors.ts
@@ -4,12 +4,14 @@ import {
   newScheduledTimestampData,
   removeAllScheduledDataByInputData,
   removeScheduledBlockData,
+  removeScheduledTimestampData,
 } from './sql/rollup_inputs.queries.js';
 import type {
   INewScheduledHeightDataParams,
   INewScheduledTimestampDataParams,
   IRemoveAllScheduledDataByInputDataParams,
   IRemoveScheduledBlockDataParams,
+  IRemoveScheduledTimestampDataParams,
 } from './sql/rollup_inputs.queries.js';
 import { strip0x } from '@paima/utils';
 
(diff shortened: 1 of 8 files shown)


packages/node-sdk/paima-sm/src/index.ts
```diff
@@ -585,8 +602,9 @@ async function processScheduledData<Events extends AppEvents>(
           {
             version: 1,
             mainChainBlochHash: strip0x(latestChainData.blockHash),
-            blockHeight: data.future_block_height,
+            blockHeight: latestChainData.blockNumber,
             prevBlockHash: prevBlockHash == null ? null : prevBlockHash.toString('hex'),
+            // TODO: should this be the actual timestamp for events scheduled by timestamp?
```
SebastienGllmt, line 607: Good question. Maybe the original timestamp can go inside `origin`?

ecioppettini:
tested this with this setup:

```ts
enum PrecompileNames {
  Default = "default",
}

const precompiles = generatePrecompiles(PrecompileNames);

async function stf(
  inputData: STFSubmittedData,
  blockHeader: PreExecutionBlockHeader,
  randomnessGenerator: Prando,
  dbConn: PoolClient
): Promise<{
  stateTransitions: SQLUpdate[];
  events: EventQueue<typeof eventDefinitions>;
}> {
  console.log("inputData", inputData);
  console.log("blockHeader", blockHeader);

  const stateTransitions = [];

  if (blockHeader.blockHeight == 3) {
    stateTransitions.push(
      createScheduledData(
        "scheduledData",
        { timestamp: new Date(3.7 * 1000) },
        { precompile: precompiles.default }
      )
    );
    stateTransitions.push(
      createScheduledData(
        "scheduledData2",
        { timestamp: new Date(3.2 * 1000) },
        { precompile: precompiles.default }
      )
    );
    stateTransitions.push(
      createScheduledData(
        "scheduledData3",
        { timestamp: new Date(4.1 * 1000) },
        { precompile: precompiles.default }
      )
    );
    stateTransitions.push(
      createScheduledData(
        "scheduledDataBlock",
        { blockHeight: 5 },
        { precompile: precompiles.default }
      )
    );
  }

  return { stateTransitions, events: [] };
}

function gameStateTransitionRouter(blockHeight: number) {
  return stf;
}

(async function main(): Promise<void> {
  spawn("docker", ["compose", "up"]);

  await new Promise((resolve) => setTimeout(resolve, 5000));

  const anvil = spawn("anvil", [
    "--port",
    "8545",
    "--chain-id",
    "31337",
    "--timestamp",
    "0",
  ]);

  process.on("exit", () => {
    anvil.kill();
  });

  await new Promise((resolve) => setTimeout(resolve, 3000));

  await setNextTimestamp(1);
  const contractAddress = deployPaimaL2Contract();

  console.log("Paima L2 contract deployed to:", contractAddress);

  runPaimaEngine(
    gameStateTransitionRouter,
    precompiles,
    {},
    {},
    { default: () => {} }
  );

  await new Promise((resolve) => setTimeout(resolve, 1000));

  await setNextTimestamp(2);
  await mineBlock();

  await setNextTimestamp(3);
  // to trigger stf at block 3 and schedule the events
  submitPaimaInput(contractAddress, Buffer.from(`hello`, "utf8"));

  await setNextTimestamp(4);
  await mineBlock();

  await setNextTimestamp(5);
  await mineBlock();
})();
```

The stf gets triggered in the order:

1. scheduledData2 (block 4)
2. scheduledData (block4)
3. scheduledData3 (block5)
4. scheduledDataBlock (block5)

note that the first 2 are reversed (the order is not by creation, but by timestamp).

packages/paima-sdk/paima-chain-types/src/types.ts
```diff
@@ -43,6 +43,11 @@ export interface SubmittedData {
     contractAddress: null | string;
     /** Name/id of the primitive that triggered this event, if known */
     primitiveName: null | string;
+
+    /* Timestamp if this is a timer scheduled with a date instead of a block
+     * number. This will be always lower or equal than the underlying block's
+     * timestamp */
+    scheduledAt: null | Date;
```
SebastienGllmt, line 50: I would specify the unit of time here (ex: `scheduleAtMs` for milliseconds

ecioppettini: isn't that already implied by the `Date` type though? And it could actually be something else, since you could call `toISOString` on it or something (instead of `getTime()`)

Or should we change this to `number` instead?

SebastienGllmt: oh my bad, I didn't notice you used `Date` as the type. Date isn't JSON compatible, so I would avoid it for types that are likely to be converted to JSON like this one, but I'm okay either way

SebastienGllmt (approved):

SebastienGllmt:
needs to be rebased

packages/node-sdk/paima-db/src/scheduled-constructors.ts
```diff
@@ -23,7 +25,7 @@ import { strip0x } from '@paima/utils';
  */
 export function createScheduledData(
```
SebastienGllmt, line 26: `deleteScheduledData` needs to also be updated to take a trigger

SebastienGllmt (approved):

PR #436, opened 2024-09-27
Migrate to loding the new Paima config system

SebastienGllmt:
This migrate from loading the old yaml file type to the new JSON object generated from the new config system


packages/node-sdk/paima-runtime/src/cde-config/loading.ts
SebastienGllmt, line 1: I only made minimal changes to this file at the moment, but it needs to be changed to read the new JSON format instead of the old yaml format

PR #437, opened 2024-09-27
Move to new STF registering style for new template

SebastienGllmt:


packages/node-sdk/paima-sm/src/types.ts
```diff
@@ -699,12 +700,13 @@ export type GameStateTransitionFunctionRouter<Events extends AppEvents> = (
   blockHeight: number
 ) => GameStateTransitionFunction<Events>;
 
-export type GameStateTransitionFunction<Events extends AppEvents> = (
-  inputData: STFSubmittedData,
-  blockHeader: PreExecutionBlockHeader,
-  randomnessGenerator: any,
-  DBConn: PoolClient
-) => Promise<{
+export type BaseStfInput = {
```
SebastienGllmt, line 703: Currently unused, but I need to change `GameStateTransitionFunction` to use it (it's a breaking change though)

PR #438, opened 2024-10-08
Fix some imports under Node's ESM mode

SpaceManiac:
Since `paima-engine` is now usable as a Node library and not just as a pre-ESBuild'd bundle, some import syntax that works in ESBuild but not in raw Node needs to be updated.

- Fix imports from CommonJS `web3-utils`, `js-sha3`, and `pg`
  - Only fields slurped by Node 20 / `cjs-module-lexer` 1.3.1 can be ESM imported, the others become fields on `.default`
- Update `@dcspark/cip34-js` to `3.0.1`
- Use `.default` on the dynamic `mina-signer` import
  - [Per MDN][import], `import()` expressions always resolve to a module object, which will have the exports (including `default`) as fields, so this is logically correct
  - This is needed for it to compile (pass tsc) on my system
  - Others have reported it only works the other way, so this may need more investigation
- Use `await import()` instead of `require()`
  - NOTE: This adds one top-level await. Middleware esbuilds might need `format: 'esm'`.
  - Hacky replacement to put everything in global scope like before: `Object.assign(window, (await import('./middleware.js')).default)`
  - Or, upcoming Paima 2 work may end up removing this path entirely
- Use `import.meta.dirname` instead of `__dirname`
- Add `/index.js` or `.js` to some imports

Also:
- Enable source maps in tsconfig

Previously: #367

[import]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import


packages/node-sdk/paima-funnel/src/cde/cardanoPool.ts
```diff
@@ -1,7 +1,7 @@
 import type { CdeCardanoPoolDatum, ChainDataExtensionCardanoDelegation } from '@paima/sm';
 import { ChainDataExtensionDatumType, DEFAULT_FUNNEL_TIMEOUT, timeout } from '@paima/utils';
-import { Routes, query } from '@dcspark/carp-client/client/src';
-import type { DelegationForPoolResponse } from '@dcspark/carp-client/shared/models/DelegationForPool';
+import { Routes, query } from '@dcspark/carp-client/client/src/index.js';
```
ecioppettini, line 3: I think this should work with just `@dcspark/carp-client` ?

packages/paima-sdk/paima-utils/src/config.ts
```diff
@@ -3,7 +3,7 @@ import type { VersionString } from './types/index.js';
 // https://github.com/flexdinesh/browser-or-node/blob/master/src/index.ts
 const isNode: boolean =
   typeof process !== 'undefined' && process.versions != null && process.versions.node != null;
-if (isNode) {
+if (isNode && typeof require !== 'undefined') {
```
ecioppettini, line 6: why is it fine to skip this? isn't the point of this to be able to run thing in node and with esm packages at the same time?

tsconfig.base.json
```diff
@@ -26,6 +26,7 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitOverride": true,
     "skipLibCheck": true,
+    "sourceMap": true,
```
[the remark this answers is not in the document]
SpaceManiac: It's not necessary for things to work, but it improves stack traces

packages/paima-sdk/paima-utils/src/config.ts
SebastienGllmt: yeah I added this piece of code a while back as a temporary hack to fix another issue I was having, and then left it in because I figured it couldn't hurt. It's maybe not a bad idea to delete it entirely

SpaceManiac: So far this PR tweaks the imports but doesn't materially change functionality. I could trash this block if you wanted, or we can do it later (maybe in a separate PR, or in the v2 branch).

SebastienGllmt: yeah I can do this in v2, so no need to bother if you're not blocked on this

ecioppettini (approved):
LGTM

PR #439, opened 2024-10-26
Make some APIs more flexible and fix corner-case encoding bugs

SpaceManiac:
- Fix possible crash in presync on second run due to passing network name where caip2 is expected.
- Allow `getRelatedWallets` to accept any `IDatabaseConnection`, not just `PoolClient`.
- Make ERC721 primitive `scheduledPrefix` optional in case a game doesn't need to react to mint events.
- Fix middleware `buildQuery` helper not escaping values; e.g. `{"a": "b&c=d"}` now properly round-trips.
- Fix snapshots not saving if DB password contains special characters.
- Fix `getAdjustedHeight` for emulated block mode always infinite looping.
- Use `toLowerCase` instead of `toLocaleLowerCase` on wallet addresses to avoid theoretical problems if a server is ever run in [Turkish locale](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase#description).
- Attempt to fix inconsistent CJS/ESM incompatibility issues with `mina-signer` by upgrading to a version that has native ESM support.
- Tweak setAchievementProgress to keep existing completed_date if set.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -514,6 +514,21 @@
         "url": "https://github.com/sponsors/isaacs"
       }
     },
+    "node_modules/@aiken-lang/aiken/node_modules/prettier": {
+      "version": "3.2.5",
+      "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.2.5.tgz",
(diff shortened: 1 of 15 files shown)


acedward:
We have moved to Effectstream rewrite.
We need to check if any of these are still an issue.

PR #440, opened 2025-06-04
Feature/tui

acedward:


packages/build-tools/collector/src/http-server.ts
```diff
@@ -141,17 +137,14 @@ server.post("/v1/logs", async (request: any, reply: any) => {
...
+
+          exportData({
+            component: scope.name,
+            namespace: namespace,
+            level: (logRecord.severityNumber as unknown as SeverityNumber) ??
               SeverityNumber.UNSPECIFIED,
-            (log) => {
-              const target = JSON.parse(otelStringify(logRecord.body));
-              // console.log(2,4) gets turned into an array [2,4] so we re-spread it)
-              return Array.isArray(target) ? log(...target) : log(target);
```
SebastienGllmt, line 152: was it okay to drop these two lines?

packages/build-tools/orchestrator/src/logging.ts
```diff
@@ -51,24 +62,45 @@ const decoder = new TextDecoder();
...
-export const rawLogHandler: LogHandler = (chunk, source, component, namespace) => {
-  Deno[source].write(chunk);
+export const rawLogHandler: LogHandler = (
+  chunk,
+  source,
+  component,
+  namespace,
+) => {
+  if (currentOutput === "default") {
+    // Deno[source].write(chunk);
```
SebastienGllmt: TODO?

packages/build-tools/orchestrator/src/process.ts
```diff
@@ -52,7 +63,13 @@ export const $ = (params: {
   }).spawn();
   process.ref(); // wait until all child processes die before killing parent
 
-  processes.push(process);
+  const processComponent = {
+    process,
+    args: params.args,
+    alive: true,
+    component: params.component ?? "unknown",
```
SebastienGllmt: It's not type-safe to set this to "unknown"?

packages/build-tools/tui/deno.json
```diff
@@ -3,25 +3,29 @@
   "version": "0.3.0",
   "exports": "./src/mod.ts",
   "tasks": {
-    "dev": "deno run --unstable-node-globals -A --check src/main.ts"
+    "check": "echo 'no check'",
```
SebastienGllmt: probably you meant to remove this

packages/build-tools/collector/src/system-exporters/api-stream.ts
```diff
@@ -0,0 +1,66 @@
...
+const MAX_DATA_ITEMS = 1000;
+
+const dataStore: TsLogExported[] = [];
+
+export function addData(item: TsLogExported) {
+  dataStore.push(item);
+  // Keep only the latest 1000 items
+  if (dataStore.length > MAX_DATA_ITEMS) {
+    dataStore.splice(0, dataStore.length - MAX_DATA_ITEMS);
+  }
```
SebastienGllmt: this feels slow (it's O(n) per time `addData` is called when we've reached the max number of elements). I think this will be way faster if you instead use a ring buffer library where you can just initialize the ring buffer to a size of `MAX_DATA_ITEMS`

packages/build-tools/tui/src/hooks/useLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+    return () => {
+      clearInterval(timer);
+    };
+  }, [write, hasShownConnectionError]);
+};
+
+// Helper function to convert severity numbers to readable names
+function getSeverityName(severity: number): string {
+  switch (severity) {
+    case 0:
```
SebastienGllmt: you can use `DefaultLogLevels` for this enum if you want

packages/build-tools/collector/src/exporters.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// This file defines the exporters pipelines used by the collector to export the data.
+export type ExportData = {
+  component: string;
+  namespace: Namespace;
+  level: SeverityNumber;
+  message: string | string[];
+};
+
+const exporters: {
```
SebastienGllmt: do we still need this at all? On one hand, it can't hurt. On the other hand, it increases code complexity without even a comment about why it's here

packages/build-tools/orchestrator/src/process.ts
```diff
@@ -3,6 +3,14 @@ import type { Namespace } from "@paima/log";
 import { ComponentNames } from "@paima/log";
 import type { ValueOf } from "@paima/utils";
 
+export type ProcessComponent = {
+  process: Deno.ChildProcess;
+  component: ValueOf<typeof ComponentNames>;
+  args: string[];
+  alive: boolean;
```
SebastienGllmt, line 17: we probably need a comment somewhere describing about how we want to make that a child process failing causes all of Paima to fail EXCEPT if the user is explicitly restarting the service (where Paima can make sure you're only restarting things in ways that don't break things), and how we're making this work

packages/build-tools/collector/src/exporters.ts
acedward: I will revert/remove this file, now that I merged the TUI logs through the [tsLog additional transports.](https://tslog.js.org/#/?id=simple-transport-example)
Eventually we will need something like this to hook to [export to other backends](https://opentelemetry.io/ecosystem/registry/?component=exporter&language=js)

SebastienGllmt: keep in mind this collector is mostly just for developer environments. For a proper production environment where using docker is fine, you wouldn't use this JS collector and you would run a real opentelemetry collector in docker. The only reason I built this collector JS project is because I didn't want to require docker just to collect logs

packages/build-tools/orchestrator/src/http-server.ts
```diff
@@ -0,0 +1,59 @@
...
+  "SHELL",
+  "STORE_HISTORICAL_GAME_INPUTS",
+];
+server.get("/setup", function handler() {
+  const obj: Record<string, string> = {};
+  env.forEach((e) => {
+    if (typeof e === "string") {
+      obj[e] = Deno.env.get(e) ?? "undefined";
+    } else {
+      obj[e.name] = e.isSecret
```
SebastienGllmt, line 42: tslog comes with a secret masking feature. Wonder if we should leverage this here as well (or at least register the secrets maybe?)

acedward: (I just left this until we have a proper env variable settings loading)

Sounds good idea.
Looks that passing `maskValuesOfKeys` with every known secret value, they will be masked in the logs.

packages/build-tools/collector/src/system-exporters/api-stream.ts
```diff
@@ -0,0 +1,118 @@
...
+const PORT = Deno.env.get("COLLECTOR_LOG_PORT")
+  ? Number(Deno.env.get("COLLECTOR_LOG_PORT"))
+  : 11033;
+
+const server = fastify();
+
+server.get("/v1/data", async (request: any, reply: any) => {
+  const copy = getData();
+  reply.status(200).send(copy);
+});
```
SebastienGllmt: I feel like we probably at the least need a comment explaining what's going on here. I think the fact this is confusing is my fault

In OpenTelemtry, there are three steps:

```
app   ---receiver--->   collector   ---exporter--->   service
```

The collector is optional. It's meant to help decouple the app from the service to allow things like multiplexing, aggregation of multiple apps, etc.

Since all of opentelemetry requires Go and Docker, I thought this was a bit too complicated for debugging purposes so I wrote my own JS code that collects logs from all the different services and prints them to console. You can think of this as combining the collector and exporter together into a single thing

The thing though is that, in production, you wouldn't run this in-memory collector and you'd have a real collector as part of your docker setup. I felt this was probably fine because in production you probably don't care as much about the console logs, and would could instead be logging at logs through a more dedicated system ex: Prometheus)

Having this `v1/data` endpoint here is kind of weird though, because a real connector won't have this endpoint. We're saying that `LogsViewer` is also not a *real* opentelemetry service, and you would only be able to use it in dev mode (as opposed to as an alternative / in parallel to Prometheus). 

If we're okay with going that route, I don't think it's that big of a deal. We should at least comment this decision somewhere though since it's not so intuitive

packages/build-tools/orchestrator/src/tmux/README.md
```diff
@@ -0,0 +1,18 @@
...
+{
+    name: string,
+    command: string,
+    split_horizontal?: boolean,
+    split_vertical?: boolean,
+}[]
+```
+If neither split_vertical or split_horizontal is defined, then the command is executed in the first pane.
+
+* It uses the `tmux.confg` as configuracion file.
```
SebastienGllmt, line 18: Probably we should start populating the `docs` folder with stuff like this to document how Paima Engine v2 works

packages/build-tools/tui/src/tab/ProcessesSection.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ) => (prev > 0 ? prev - 1 : processes.length - 1));
+      } else if (key.downArrow && processes.length > 0) {
+        setSelectedIndex((
+          prev,
+        ) => (prev < processes.length - 1 ? prev + 1 : 0));
+      } else if (key.return && processes.length > 0) {
+        // Enter key pressed - show confirmation
+        // const selectedProcess = processes[selectedIndex];
+        // setProcessToRestart(selectedProcess);
+        // setShowConfirmation(true);
```
SebastienGllmt: did we want to leave these here?

packages/build-tools/orchestrator/src/process.ts
acedward: I think this should be part of a next PR. As we need to decide on how/if processes can be restart (as each process has it's own rules)

packages/build-tools/collector/src/system-exporters/api-stream.ts
acedward: Fixed*, following a more otel architecture. Moving the server into the log domain, so it works as a collector/exporter

packages/build-tools/orchestrator/src/process.ts
```diff
@@ -18,41 +27,82 @@ export class AbortProcessStart extends Error {
...
+
+let shutdownCalled = false;
+export async function shutdown(exitCode: number = 0): Promise<void> {
+  if (shutdownCalled) {
+    return;
+  }
+  shutdownCalled = true;
+  // We are shutting down the processes, so redirect logs to the stdout
+  // So we see the shutdown messages.
+  setCurrentOutput("stdout");
```
SebastienGllmt: won't this mean that if Paima ever crashes, the logs (which presumably contain the crash logs inside it) will only ever be written to stdout and not to otel/log files?

packages/build-tools/orchestrator/src/process.ts
```diff
@@ -18,41 +27,82 @@ export class AbortProcessStart extends Error {
...
-}): Deno.ChildProcess => {
+  stdin?: "inherit" | "piped" | "null" | undefined;
+  stdout?: "inherit" | "piped" | "null" | undefined;
+  stderr?: "inherit" | "piped" | "null" | undefined;
+}): ProcessComponent => {
   if (failed) {
     throw new AbortProcessStart("Shutdown already called");
   }
-  const process = new Deno.Command("deno", {
+  const abortController = new AbortController();
```
SebastienGllmt: Why are we now using a separate abort controller for every process? The point of the `AbortControllers` was so that we could easily abort all processes in a group by having them all share the same signal (ex: just send a one signal to reset all DB-related processes)

Presumably you made this change so that we could shutdown processes individually in the TUI, but as I mentioned we actually can't shutdown processes individual because it could get Paima into a bad state. We should instead define process groups that shut down together, and that's what these shared signals are for

packages/build-tools/tui/src/logs-standalone.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    // Write output to file.
+    for (const [namespace, buffer] of Object.entries(fileBuffer)) {
+      await Deno.writeFile(
+        `./logs/${namespace}-${ddmmyyyy}.log`,
+        new TextEncoder().encode(buffer.join("\n") + "\n"),
+        { append: true },
+      );
+    }
```
SebastienGllmt: Do we want to write logs to a file instead of in-memory even in debug mode? Maybe the answer is yes, but I could also imagine the user accidentally using `console.log` in a loop and next they know they have a log file gigabytes in size

packages/build-tools/collector/src/http-server.ts
```diff
@@ -37,9 +38,40 @@ function printError(e: any, namespace: string, request: any) {
...
+      Array.isArray(data.message) ? log(...data.message) : log(data.message);
+    },
+  );
+}
+
+// Add a tsLog hook to export the logs to the API stream
+const API_LOG_URL = Deno.env.get("API_LOG_URL") ?? "http://localhost:11033";
+attachTransport(async (logObj) => {
+  try {
+    await fetch(API_LOG_URL + "/v1/data", {
```
SebastienGllmt: Should this not be `v1/logs` to match the opentelemetry standard? Keep in mind that to get the LogsViewer to also work in production (where there is a real opentelemetry system), the endpoint path & format has to match what opentelemetry expects

acedward: Perfect, In production we will spin a otel server and setup something like this:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

processors:
  batch:

exporters:
  otlphttp/logs:
    endpoint: "0.0.0.0:11033/v1/logs"
    tls:
      insecure: true  # Set to false and configure certificates for HTTPS

service:
  pipelines:
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlphttp/logs]
 ``` 

I need flip something around, now the paima-log write to the log backend, I will make the otel write to the log backend, and it uses paima-log to add colors.

packages/build-tools/tui/src/logs-standalone.ts
acedward: This was a quick, bad, solution. Just to be able to read the logs per-output - and only for development.
In production we should setup the otel with 
```yaml
exporters:
  file/rotated_logs:
    path: /data/collector/otlp_logs.json
    rotation:
      max_megabytes: 10
      max_days: 14
      max_backups: 500
      localtime: false
    format: json
    flush_interval: 5s
 ```

acedward: Perhaps I should use some library to limit/rotate logs here.

packages/build-tools/orchestrator/src/process.ts
acedward: In development
case 1: **I press cmd+c** while looking at the logs in our tmux system. 
This cmd closes the logs and kill the tmux, now If any log is sent to the otel, it's never "printed" on screen (as the tui is dead) - so this was a way to redirect to your terminal to visualize the logs 

case 2: **Some random process fails**, we start shutting down processes, we could close the tmux / otel at the end to see most of the process logs shutting down, but - once the tmux is killed the lines that where printed there where lost (and you are "redirected" to your original terminal) not seeing what the issue was.

In production
case 3: **Some random process fails**, we will lose logs. 
Perhaps only doing this if NODE_ENV=development?


I think would be best in to output to both outputs, 
`setCurrentOutput(["otel", "stdout"]);`
So it works well in dev and not lose logs

packages/build-tools/orchestrator/src/tmux/README.md
acedward: I created a issue to track/do this.
should we try to keep docusaurus? 
I gave it a quick try with deno, but didn't work for me. 
~~We also have some alternatives that as https://github.com/getzola/zola (it's distributed as single binary, and there is npm:zola-bin)~~ edit: it doesn't look like it has to many doc features, more generic static site builder, that can be used for docs

packages/build-tools/orchestrator/src/process.ts
SebastienGllmt: note the old code I wrote purposely tried to shutdown the otel stuff last to keep it alive to print as long as possible (and I guess in our case, we would make that apply to tmux as well)

packages/build-tools/tui/src/logs-standalone.ts
SebastienGllmt: I think this is probably more of an issue in debug builds that production, because in debug builds you're more likely to have a really spammy `console.log` to test something (which is when you accidentally write gigabytes of logs)

packages/build-tools/collector/src/http-server.ts
acedward: https://github.com/PaimaStudios/paima-engine/pull/440/commits/211ebcae46b05f2bbed84545a1ecb9045c101523

So now the Tui/Logs are (almost) compatible with the setup above.
We just to adjust our implementation of the otel collector exporter to match the output of otlphttp/logs

Added log rotation/max size through tsLogs

packages/build-tools/tui/src/logs-standalone.ts
acedward: https://github.com/PaimaStudios/paima-engine/commit/211ebcae46b05f2bbed84545a1ecb9045c101523
now using tsLogs to solve this

packages/build-tools/orchestrator/src/process.ts
acedward: Merged Abort Controllers 

https://github.com/PaimaStudios/paima-engine/pull/440/commits/03755ff30bcd4327dbca7b56c6446d0c33accaf5

packages/build-tools/orchestrator/src/process.ts
acedward: Fixed https://github.com/PaimaStudios/paima-engine/pull/440/commits/03755ff30bcd4327dbca7b56c6446d0c33accaf5

SebastienGllmt:
although what we've built isn't quite otel-compatible yet, I'm merging this PR anyway because it's gotten fairly big and it is functional for our current purposes

PR #448, opened 2025-07-10
Feature/v next scheduled inputs

acedward:
* Process blocks scheduled by height
* Process blocks scheduled by date
* Fixed some issues caused by deleting/recreating the evm json address file - now lazy loading contents
* Refactor now Processed Primitives write Scheduled data for the same block-height, and then executed in the State Machine, this will allow to parallel process in the future.
* New tests, and made test more consistent.
* Added example for input that creates scheduled data.
* Added scheduled data execution result into database


example/shared/data-types/src/grammar.ts
```diff
@@ -5,6 +5,7 @@ import { localhostConfig } from "./config.ts";
 export const grammar = {
   schedule: [
     ["tick", Type.Integer()],
+    ["type", Type.String()], // Type.Enum(Type.Literal("block"), Type.Literal("timestamp"))],
```
SebastienGllmt, line 8: Did you mean to leave this as a comment?

packages/node-sdk/runtime/src/process-blocks.ts
```diff
@@ -194,8 +197,76 @@ export function* processFinalizedBlock(
...
+    );
+    /* STEP 4.b: Extract the scheduled data by timestamp. */
+    const scheduledData2 = yield* call(() =>
+      getFutureGameInputByMaxTimestamp.run({
+        max_timestamp: new Date(value.timestamp),
+      }, dbConn)
+    );
+
+    /* STEP 5: Process the scheduled data in the State Machine. */
+    // TODO What should be the order of the scheduled data - per id?
```
SebastienGllmt, line 214: Sort by `rollup_inputs` ID I guess

SebastienGllmt (approved):

packages/node-sdk/runtime/src/process-blocks.ts
acedward: Issue PAI-64

PR #453, opened 2025-07-18
State Machine Generator

acedward:
State Machine States are now Generator Functions


packages/node-sdk/sm/verify-signature.ts
```diff
@@ -35,14 +35,11 @@ export function* verifySignature(
       // IMPORATNT: sync generator cannot resolve promises.
```
acedward:
```suggestion
      // IMPORTANT: sync generator cannot resolve promises.
```

packages/node-sdk/runtime/src/process-blocks.ts
```diff
@@ -204,29 +204,29 @@ export function* processFinalizedBlock(
...
+          chain: {
+            blockNumber: value.blockNumber,
+            transactionHash: "0x0",
+          },
+        };
+        const op = gameStateTransitions(
+          value.blockNumber,
+          input,
+        );
+        yield* executeGeneratorStepByStep(op, dbConn);
```
acedward:
```suggestion
        const gameSTFGenerator = gameStateTransitions(
          value.blockNumber,
          input,
        );
        yield* executeGeneratorStepByStep(gameSTFGenerator, dbConn);
```

packages/node-sdk/runtime/src/process-blocks.ts
```diff
@@ -204,7 +204,8 @@ export function* processFinalizedBlock(
     let index_in_block = 0;
     if (gameStateTransitions && scheduledData.length > 0) {
       for (const data of scheduledData) {
-        const { status } = yield* until(tryOrRollback(dbConn, async () => {
```
SebastienGllmt, line 207: are you sure about removing `tryOrRollback` here? It feels like this will introduce a bug where a single failed transition will cause the entire block to fail

PR #467, opened 2025-08-07
pagination paima endpoints

Yiakman:
# 🚨 BREAKING CHANGE: Cursor-Based Pagination Implementation

## 📋 Overview

This PR refactors the pagination system for all array-returning endpoints,
replacing the previous `limit/offset` strategy with a more performant and robust
**cursor-based** (keyset) pagination. This introduces a **breaking change** in
both the query parameters and the response format for multiple endpoints.

### **Response Format Change**

All array-returning endpoints now use a **paginated response wrapper** instead
of returning arrays directly.

#### **Before (v1.x)**

```json
[
  { "address": "0x123...", "account_id": 1 },
  { "address": "0x456...", "account_id": 2 }
]
```

#### **After (v2.x)**

```json
{
  "data": [
    { "address": "0x123...", "account_id": 1 },
    { "address": "0x456...", "account_id": 2 }
  ],
  "pagination": {
    "limit": 2,
    "hasMore": true,
    "nextCursor": "eyJhY2NvdW50X2lkIjoyLCJhZGRyZXNzIjoiMHg0NTY..."
  }
}
```

### **Affected Endpoints**

The following **Paima Engine Runtime API** endpoints now return paginated
responses:

#### **🔍 Paima Engine Runtime API**

- `GET /addresses` - User addresses and account mappings
- `GET /scheduled-data` - Scheduled rollup inputs
- `GET /tables/:tableName` - Generic table data (can be large)
- `GET /primitives/:primitiveName` - Primitive aggregated data

#### **🛠️ Build Tools (Selective)**

- `GET /v1/data` - TUI log server data (can have many log entries)

### **Pagination Parameters**

All paginated endpoints now support the following optional query parameters:

- `limit` (optional): Number of items to return per page (default: 20, max:
  1000).
- `after` (optional): An opaque cursor string returned from a previous request's
  `nextCursor` field. When provided, it fetches the next page of results. If
  omitted, it returns the first page. It uses a base64 encoded JSON.

```bash
# Get the first page (default limit)
GET /addresses

# Get the first 10 items
GET /addresses?limit=10

# Get the next page using the cursor from the previous response
GET /addresses?limit=10&after=eyJhY2NvdW50X2lkIjoyLCJhZGRyZXNzIjoiMHg0NTY...
```

### **Performance and Consistency**

- **Data Consistency**: Unlike offset-based methods, this approach prevents
  items from being skipped or duplicated if the underlying data changes between
  requests.
- **No Total Count**: The `count` parameter and `total` field in the response
  have been removed. Calculating the total number of rows is an expensive
  operation that negates the performance benefits of cursor pagination. The
  `hasMore` flag should be used to determine if another page exists.
- **Backwards Compatible (First Page)**: Old queries without pagination
  parameters will still work for the first page, but the response format has
  changed.


packages/node-sdk/db/src/sql/accounts.queries.ts
```diff
@@ -15,17 +15,28 @@ export interface INewAddressQuery {
   result: INewAddressResult;
 }
 
-const newAddressIR: any = {"usedParamSet":{"address":true},"params":[{"name":"address","required":true,"transform":{"type":"scalar"},"locs":[{"a":41,"b":49}]}],"statement":"INSERT INTO addresses (address) \nVALUES (:address!)"};
+const newAddressIR: any = {
+  "usedParamSet": { "address": true },
+  "params": [{
```
Yiakman: This is just formatting done by the linter

packages/node-sdk/db/src/sql/accounts.sql
```diff
@@ -45,4 +45,9 @@ SELECT
     accounts.primary_address as "primary_address"
 FROM addresses
 LEFT JOIN accounts ON accounts.primary_address = addresses.address
-ORDER BY addresses.account_id;
+ORDER BY addresses.account_id
+LIMIT COALESCE(:limit, 1000) OFFSET COALESCE(:skip, 0);
```
Yiakman: By default pagination is always being passed with DEFAULT_PAGINATION_LIMIT 20 and MAX_PAGINATION_LIMIT 1000

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -155,20 +165,65 @@ export const startHttpServer = function* (
...
+        : {};
+
+    // Only run count query if explicitly requested
+    const addressesPromise = runPreparedQuery(
+      getAllAddresses.run(paginationParams, dbConn),
       "addresses",
     );
-    return result;
+
+    const countPromise = count
```
Yiakman: counts are calculated only if count=true param is passed and processed after the main query

packages/build-tools/explorer/client/src/components/AddressesTable.tsx
```diff
@@ -394,8 +394,9 @@ export function AddressesTable() {
         throw new Error(`HTTP error! status: ${response.status}`);
       }
 
-      const data = await response.json();
-      setAddresses(data);
+      const jsonResponse = await response.json();
```
Yiakman, line 397: Unsure on how to handle pagination inside the explorer

example/client/node/scripts/e2e.general.test.ts
```diff
@@ -393,11 +393,32 @@ export async function generalTest(db: Client, sharedState: SharedState) {
   });
 
   await assert("Check System API Table Data", async () => {
+    const limit = sharedState.paima_state_machine_counter * 2; // 2x the data length
     const response = await fetch(
-      `http://localhost:${ENV.PAIMA_API_PORT}/tables/user_state_machine`,
+      `http://localhost:${ENV.PAIMA_API_PORT}/tables/user_state_machine?count=true&limit=${limit}`,
```
SebastienGllmt: I didn't read too much into this PR so maybe this is safe, but usually for blockchain related stuff you need some kind of anchor for pagination because new data could come in between queries (and missing data between pages because things shifted is often a critical failure)

packages/build-tools/tui/src/logs-server.ts
```diff
@@ -79,9 +85,18 @@ export class LogServer {
...
 
-    this.server.get("/v1/data", (request: any, reply: any) => {
+    this.server.get("/v1/data", {
+      schema: {
+        querystring: PaginationQuerySchema,
+        response: {
+          200: createPaginatedResponseSchema(OTelLogSchema),
+        },
+      },
+    }, (request: any, reply: any) => {
```
acedward: remove any

PR #473, opened 2025-08-11
filter queries to system and dynamic tables

Yiakman:


packages/node-sdk/runtime/src/api/table-filter.ts
```diff
@@ -0,0 +1,81 @@
+import type { AllSyncProtocols } from "@paima/sync";
+import { getPrimitiveIntermediatePrefix, getPrimitivePrefix } from "@paima/db";
+
+// System tables created by Paima core migrations (see db/migrations/up.sql)
+const SYSTEM_TABLES = new Set<string>([
```
SebastienGllmt, line 5: seems a bit dangerous to have this as a hardcoded list since it will definitely get out of sync

Yiakman:
Closed in favor of [this solution instead
](https://github.com/PaimaStudios/paima-engine/pull/486)

PR #478, opened 2025-08-13
Replay protection nonce rules in v-next

Yiakman:


e2e/client/node/scripts/e2e.general.test.ts
```diff
@@ -267,6 +274,50 @@ export async function generalTest(db: Client, sharedState: SharedState) {
     },
   );
 
+  // Capture current nonce count after the first valid batched submission
+  const initialNonceCountResult = await db.query(
+    `SELECT COUNT(*)::int as c FROM public.nonces;`,
```
acedward, line 273: We use PGTyped for all queries

e2e/client/node/scripts/e2e.general.test.ts
```diff
@@ -307,17 +358,117 @@ export async function generalTest(db: Client, sharedState: SharedState) {
...
+  await assertSQL<{ c: number }>(
+    "Nonces table accessible",
     db,
-    `SELECT * FROM public.nonces;`,
-    (res) => res.rows.length === nonce_counter,
-    (res) => {
-      return res.rows.length === nonce_counter;
-    },
+    `SELECT COUNT(*)::int as c FROM public.nonces;`,
+    (res) => res.rows.length === 1,
```
acedward: Hmm this number should be higher now, as direct TX also increment the number of nounces

e2e/client/node/scripts/e2e.general.test.ts
```diff
@@ -307,17 +358,117 @@ export async function generalTest(db: Client, sharedState: SharedState) {
...
+      });
+      const tx2 = await walletClient.writeContract({
+        account,
+        chain: hardhat,
+        address: addr,
+        abi,
+        functionName: "paimaSubmitGameInput",
+        args: [argData],
+        value: parseEther("0.0000000001"),
+      });
```
acedward: Please use e2e-contracts to write, so we can centralize the system state

acedward: maybe submitGameInput add a flag so it can skip editing the sharedState

`submitGameInput(input, privateKey, updateState = true) {...}`

packages/node-sdk/sm/primitives/evm/rpc/paima-l2.ts
```diff
@@ -251,10 +277,26 @@ export default function* processPaimaL2SyncProtocolResponse(
...
+          ? Value.Decode(
+            TypeboxHelpers.Evm.Address,
+            outerLayerData.userAddress,
+          )
+          : outerLayerData.userAddress;
+      const toHash = String(paima_block_height) + userAddress + gameInputStr;
+      directNonce = keccak256(stringToHex(toHash));
+    } catch {
+      /* ignore decode failure; skip dedup for this input */
+    }
```
acedward: In the block processing we cannot have unknown states, let it throw an exception if it must, but it should never happen (remove the try catch, and if there is a specific condition that fails, let's capture and manage it)

packages/node-sdk/sm/primitives/evm/rpc/paima-l2.ts
```diff
@@ -185,6 +184,14 @@ export default function* processPaimaL2SyncProtocolResponse(
     PrimitiveEvmRpcPaimaL2Payload,
     response.output.payload,
   );
+  // Fetch the block timestamp (ms) for 24h validation
+  const [blockInfo] = yield* World.resolve(getBlockHeights, {
+    block_heights: [paima_block_height],
+  });
+  const blockTimestampMs = blockInfo?.ms_timestamp
+    ? (blockInfo.ms_timestamp as Date).getTime()
+    : undefined;
```
acedward: This should just fail if is not available

```suggestion
  const blockTimestampMs = blockInfo!.ms_timestamp.getTime();
```

packages/node-sdk/sm/primitives/evm/rpc/paima-l2.ts
```diff
@@ -199,6 +206,25 @@ export default function* processPaimaL2SyncProtocolResponse(
       const { parsed } = batchedMessage;
       const { userAddress, millisecondTimestamp, userSignature, gameInput } =
         parsed;
+      // 24h timestamp validation for batched inputs
+      if (blockTimestampMs !== undefined) {
```
acedward: cannot be undefined

packages/node-sdk/sm/primitives/evm/rpc/paima-l2.ts
```diff
@@ -199,6 +206,25 @@ export default function* processPaimaL2SyncProtocolResponse(
...
+        const signedTs = Number(millisecondTimestamp);
+        if (
+          !Number.isFinite(signedTs) ||
```
acedward: if you want we can throw an error if `!Number.isFinite(signedTs)` but not add as condition to the logic

docs/site/docs/home/1000-paima-engine/1090-security/1090-duplicate-protection.md
```diff
@@ -0,0 +1,97 @@
+---
+id: duplicate-protection
+title: Duplicate Protection in Paima L2
+slug: /paima-engine/duplicate-protection
+sidebar_label: Duplicate Protection
+---
+
+# Replay Protection in Paima L2 (v2)
+
```
acedward: Can you leave the Section called "Security"
So the side bar looks simpler to understand 🙏

e2e/client/node/scripts/e2e.general.test.ts
```diff
@@ -267,6 +268,50 @@ export async function generalTest(db: Client, sharedState: SharedState) {
     },
   );
 
+  // Capture current nonce count after the first valid batched submission
+  const initialNonceCountResult = await db.query(
+    `SELECT COUNT(*)::int as c FROM public.nonces;`,
+  );
```
acedward, line 274: Important, to do db.query you must wrap in
```
let result;
try {
      await fetch(
        `http://localhost:${ENV.PAIMA_API_PORT}/db_acquire_lock?name=${testName}`,
      );
      didLock = true;
      result = await db.query(sql);
    } finally {
      if (didLock) {
        await fetch(
          `http://localhost:${ENV.PAIMA_API_PORT}/db_release_lock?name=${testName}`,
        );
      }
    }
```

docs/site/docs/home/1000-paima-engine/1090-security/1090-duplicate-protection.md
Yiakman: Left it as architecture/security or do you prefer security at the base?

e2e/client/node/scripts/e2e.general.test.ts
acedward: Let's fix in new version

docs/site/docs/home/1000-paima-engine/1090-security/1090-duplicate-protection.md
```diff
@@ -0,0 +1,97 @@
...
+- Direct:
+  - Call the contract with your input hex.
+  - Dedup is same-block: identical inputs in the same block dedup; across blocks
+    they're distinct.
+
+- Scheduled:
+  - No nonce required; engine prevents replays via atomic execution.
+
+Reference (v1 background):
+[Replay protection](https://docs.paimastudios.com/home/state-machine/direct-write/replay-protection/).
```
acedward:
```suggestion

```

PR #493, opened 2025-08-26
add schema separation logic to migration process

Yiakman:
This PR incorporateds the schema separation logic into the migration code that was present in v-next branch.

Includes an update to http-server /tables/:table endpoint that verifies that the queried table belongs to the public schema before calculating other fields

Updated readme for generate assets.ts for more clarity


packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -363,6 +364,12 @@ export const startHttpServer = function* (
       reply,
     ) => {
       const { tableName } = request.params;
+      if (tableName === "paima_engine_version_history") {
+        return reply.status(400).send({ error: "Invalid table name" });
+      }
+      if (tableName === "paima_engine_migration_history") {
+        return reply.status(400).send({ error: "Invalid table name" });
+      }
```
acedward: @Yiakman why are these special names? They are in the paima schema

Yiakman: I removed them, they remained from a test I was doing from a crash that was happening before

acedward (approved):

PR #501, opened 2025-08-29
@paima/wallet 

acedward:
# Added package for connecting web wallets
* Algorand
* Avail
* Cardano
* Evm Injected
* Evm Ethers
* (NYI) Midnight
* Mina
* Polkadot

# Pending Tasks for different PRs. Added to issue racker

- [ ] Type checking of the concise input for Wallet Methods (e.g., sendTransaction)
- [ ] Signing and Validation for other wallets than EVM
- [ ] Some injected wallets don't want to work as expected
- [ ] Implement a Local Wallet as default wallet to select
- [ ] Implement Midnight Injected (Implement midnight local?)
- [ ] Convert Ethers to Viem
- [ ] Dynamic imports for wallets
- [ ] Implement the namespace/app name in the message that are signed (?)


packages/paima-sdk/crypto/src/chains/algorand.ts
```diff
@@ -6,6 +6,34 @@ import {
...
 import type { IVerify } from "../IVerify.ts";
 import { Value } from "@sinclair/typebox/value";
+import { utf8ToHex } from "web3-utils";
+import {
+  type Transaction,
+  type SuggestedParams,
+  makePaymentTxnWithSuggestedParamsFromObject,
+  decodeSignedTransaction,
+  type SignedTransaction,
+} from "algosdk";
```
SebastienGllmt, line 16: as a reminder, we were trying to avoid static imports for these because 99% of dApps will not use Algorand at all, and it saves a lot of app startup time to not include different chains (both because of reduction in bundled file size, but also because some of these libraries initialize a bunch of WASM stuff / crypto stuff on load)

acedward: Note: there are some improvements in the next PRs for the dependencies (but still not dynamic) we first want to make it working completely and bundlable

PR #502, opened 2025-08-29
safe pagination and hybrid offset strategy

Yiakman:
Ensure no SQL injections happen to the /primitives/:primitive and /table/:table endpoint
improve pagination with a mixed offset that can be a base64 jsdn or a number for offset


packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -36,8 +37,22 @@ import {
...
   PaginationQuerySchema,
   type TypePaginationQuerySchema,
+  validateColumnName,
 } from "./pagination.ts";
 
+// Utility functions for SQL injection prevention
+function escapeColumnName(columnName: string): string {
+  // Double-quote column names to prevent SQL injection
+  return `"${columnName.replace(/"/g, '""')}"`;
+}
```
acedward: Let's move this to pagination (so we keep all SQL stuff there, the HTTP files should only have the logic regarding the server/request data parsing)

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -36,8 +37,22 @@ import {
...
+
+function validateAndEscapeColumnName(columnName: string): string | null {
+  if (!validateColumnName(columnName)) {
+    return null;
```
acedward: this should probably through an error? I see this is not handled.

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -297,7 +312,21 @@ export const startHttpServer = function* (
...
+              "Cursor for next page (base64-encoded JSON object with primary key values)",
+            examples: ["eyJhY2NvdW50X2lkIjoxMjMsImFkZHJlc3MiOiIwMXQzIn0="],
+          }),
+          Type.Number({
+            description: "Offset for pagination (0-based)",
+            minimum: 0,
+            examples: [0, 100, 500],
+          }),
+        ])),
+      }),
```
acedward, line 324: This looks triplicated?

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -590,17 +648,50 @@ export const startHttpServer = function* (
...
+            const cursorKeys = Object.keys(after);
+            const missingKeys = pkColumns.filter((pk) =>
+              !cursorKeys.includes(pk)
+            );
+            const extraKeys = cursorKeys.filter((k) => !pkColumns.includes(k));
+
+            if (missingKeys.length > 0 || extraKeys.length > 0) {
+              return reply.status(400).send({
+                error: `Cursor must contain exactly the primary key columns: ${
+                  pkColumns.join(", ")
```
acedward: can we move this to a function in pagination, it's not regarding the server - but how to build the dynamic query

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -613,7 +704,16 @@ export const startHttpServer = function* (
           console.warn(
             `[Paima Engine] WARNING: Table "${safeTableName}" has no primary key. Falling back to less performant OFFSET-based pagination.`,
           );
-          const offset = Number(parseInt(request.query.after ?? "0"));
+
+          // Use offset from pagination parameters, defaulting to 0
+          const paginationOffset = offset ?? 0;
+
+          // Validate offset is non-negative
+          if (paginationOffset < 0) {
```
acedward: This error should be handled by getPaginationParams., perhaps throw an error if < 0

packages/node-sdk/runtime/src/api/http-server.ts
```diff
@@ -765,19 +896,59 @@ export const startHttpServer = function* (
...
+                "Invalid cursor: must contain exactly one primary key field",
+            });
+          }
+
+          primaryKey = keys[0];
+          const escapedKey = validateAndEscapeColumnName(primaryKey);
+          if (!escapedKey) {
+            return reply.status(400).send({
+              error: "Invalid primary key column name in cursor",
+            });
```
acedward: idem move to pagination, as it's not http-server related

packages/node-sdk/runtime/src/api/pagination.ts
```diff
@@ -54,34 +92,57 @@ export type PaginatedResponse<T> = {
...
+          !validateCursorStructure(parsed, expectedCursorFields)
+        ) {
+          throw new Error("Invalid cursor structure");
+        }
+
+        after = parsed as T;
+      } catch (e) {
+        // Invalid cursor or offset, treat as no cursor
+        after = undefined;
+        offset = undefined;
```
acedward: if invalid, let's just pass the error and no try to fix it

PR #503, opened 2025-09-02
Feature/improve block fetch frontend

Yiakman:
Fetches blocks in between the synced_paged number returned from the latest /block-heights query and the previous one


acedward (commented):
👍

PR #508, opened 2025-09-08
Feature/avail sync

Yiakman:
Implements
- AvailClient with the calls to the node and light-client for block and data fetching
- fetcher for blocks and processing data
- App deployment: Now it starts avail node, then a deploy script registers a new appId and then it launches the light-client with that same appId so it runs on app-mode
- Updated types and primitives

Fetching data works first by checking on the block header app_lookup.index if the appId appears listed. Data is decoded from base64 in the fetcher and passed as a result.


packages/node-sdk/sm/primitives/avail/contract-state.ts
```diff
@@ -0,0 +1,51 @@
...
+    BuiltinTransitions[ConfigPrimitiveType.AvailPaimaL2]
+    .scheduledPrefix,
+    scheduledPrefix,
+    { payload },
+  );
+  yield* World.resolve(insertPrimitiveAccounting, {
+    primitive_name: response.output.syncProtocol.payload.primitiveName,
+    paima_block_height: paima_block_height,
+    payload_type: ConfigPrimitiveAccountingPayloadType.Event,
+    payload: JSON.stringify(response.output.payload) as any,
```
Yiakman: just payload

packages/node-sdk/sm/primitives/avail/contract-state.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+  if (scheduledPrefix) {
+    yield* createScheduledData(
+      JSON.stringify(scheduledInputData),
+      {
+        blockHeight: syncProtocol.payload.ownChain.blockNumber,
+      },
+      {
+        primitiveName: response.output.syncProtocol.payload.primitiveName,
+        txHash: response.output.syncProtocol.internal.transactionHash,
```
Yiakman: clean here

packages/node-sdk/sm/primitives/avail/contract-state.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { scheduledPrefix } = response.input;
+  const { payload, syncProtocol } = response.output;
+  const scheduledInputData = generateRawStmInput(
+    BuiltinTransitions[ConfigPrimitiveType.AvailPaimaL2]
+    .scheduledPrefix,
+    scheduledPrefix,
+    { payload },
+  );
+  yield* World.resolve(insertPrimitiveAccounting, {
+    primitive_name: response.output.syncProtocol.payload.primitiveName,
```
Yiakman: clean

PR #521, opened 2025-09-25
Simplify quickstart instructions and add links to check.sh

SpaceManiac:
Quickstart docs:
- Autoformat Markdown (using Prettier)
- Simplify `git clone` command

check.sh:
- Fix shellcheck warnings
- Add links to the installation instructions for Deno, Node, Forge, and Compact

---

Autoformatting can be undone if it seems noisy.


acedward (commented):
💯

PR #531, opened 2025-10-09
Fixed MQTT automatic filtering

acedward:
Fixed remoteAddres from socket for deno
Fixed localHost detection

### DIFF
--- a/packages/node-sdk/events/src/event-broker.ts
+++ b/packages/node-sdk/events/src/event-broker.ts
@@ -50,6 +50,9 @@ export class PaimaEventBroker {
       // NOTE: Using deno `client?.req?.socket?.remoteAddress` is undefined.
       //       This was defined correctly when using node.js runtime.
       // 
+      // This is a workaround for the following issue:
+      // https://github.com/denoland/deno/issues/30707
+      //
       // In https://github.com/denoland/deno/pull/20120 these new fields were added:
       // Symbol(kHandle) and Symbol(kStreamBaseField).
       //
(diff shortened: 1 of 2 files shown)


packages/node-sdk/events/src/event-broker.ts
```diff
@@ -36,15 +39,34 @@ export class PaimaEventBroker {
...
+      // @ deno we are getting `client?.req?.socket?.remoteAddress` as undefined.
 
-      // TODO We need to allow only localhost to publish.
-      //      In deno we are getting `client?.req?.socket?.remoteAddress` as undefined.
-      //
-      // if (client?.req?.socket?.remoteAddress == null)
-      //   return callback(
-      //     new Error('Error: no remove address found for  PaimaEventBroker connection')
-      // );
+      // socket has a symbol kHandle that has a symbol kStreamBaseField that has a localAddr and a remoteAddr object
```
SebastienGllmt: when doing workarounds like this, you should always link the Github issue on the upstream project that describes why we did the workaround

that way, in the future, anybody else looking at this code can get a good understanding of what the issue was, and whether or not it's been fixed in the upstream code since our code was written

acedward: I found this by debugging the socket.

Object keys that are Symbols don't appear if serialized normally, so console.logs didn't show this.

After digging into the symbols keys, the remote address was buried there. It doesn't look documented.

They added this internal field when fixing this issues with MQTT https://github.com/denoland/deno/issues/19978

SebastienGllmt: what I meant is link https://github.com/denoland/deno/issues/30707

PR #565, opened 2025-11-26
Paima Templates: Migrates Open World(renamed to world-map-2d) and fixes Minimal

Lambdanaut:


templates/minimal/packages/frontend/deno.json
```diff
@@ -0,0 +1,7 @@
+{
```
acedward, line 1: This folder was in the root, to explicitly not make it part of the deno workspace (probably you made it work)
Can you add this process to `templates/minimal/packages/client/start-script.ts` 

```
  processesToLaunch: [
    ...launchEvm("@minimal/evm-contracts"),
    { 
      name: "build-frontend",
      stopProcessAtPort: [8080],
      args: ["task", "-f", "@name-of-this-package", "build"],
      waitToExit: true,
      dependsOn: [],
    },
    { 
      name: "frontend",
      stopProcessAtPort: [8080],
      args: ["task", "-f", "@name-of-this-package", "serve"],
      waitToExit: false,
      dependsOn: ['build-frontend'],
    },
  ],
```

acedward: Nevermind, I saw you removed it from the deno workspace!

templates/world-map-2d/packages/client/node/src/state-machine/v1/transition.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// State transition when a user submits an increment
+export const submitIncrement = async (
+  player: WalletAddress,
+  blockHeight: number,
+  input: SubmitIncrementInput,
+  randomnessGenerator: Prando
+): Promise<SQLUpdate[]> => {
+  return [persistWorldCount(input.x, input.y)];
+};
```
acedward, line 55: Can we make these return `Promise<SQLUpdate>` instead (as they all return a single element)
Returning an array was a necessity of V1.
So we can also remove the for loop in the state machine file

templates/world-map-2d/packages/client/node/src/api.ts
```diff
@@ -0,0 +1,38 @@
...
+): Promise<void> {
+  // Get user stats endpoint
+  server.get("/user_stats", async (request, reply) => {
+    const { wallet } = request.query as { wallet: string };
+    if (!wallet) {
+      return reply.code(400).send({ error: "wallet parameter required" });
+    }
+
+    try {
+      const [userStats] = await getUserStats.run({ wallet }, dbConn);
```
acedward: This need to be wrapped in `runPreparedQuery(...)` 
Our In-memory DB only allows one-thread. 
As we have different process executing DB Queries

https://pglite.dev/docs/
```
As PGlite only has a single exclusive connection to the database, we provide a [multi-tab worker](https://pglite.dev/docs/multi-tab-worker) to enable sharing a PGlite instance between multiple browser tabs.
```

This function runPreparedQuery has a semaphore to allow all systems to share the single connection 💀

templates/world-map-2d/packages/client/node/deno.json
```diff
@@ -0,0 +1,46 @@
...
+    "@paimaexample/utils": "jsr:@paimaexample/utils@0.3.108",
+    "@paimaexample/sm": "jsr:@paimaexample/sm@0.3.108",
+    "@paimaexample/coroutine": "jsr:@paimaexample/coroutine@0.3.108",
+    "@paimaexample/db": "jsr:@paimaexample/db@0.3.108",
+    "fastify": "npm:fastify@^5.4.0",
+    "pg": "npm:pg@^8.14.0",
+    "@sinclair/typebox": "npm:@sinclair/typebox@^0.34.30",
+    "effection": "npm:effection@^3.5.0",
+    "@paimaexample/explorer": "npm:@paimaexample/explorer@0.3.108",
+    "@midnight-ntwrk/compact-runtime": "npm:@midnight-ntwrk/compact-runtime@^0.9.0",
```
acedward: midnight imports should not be needed 👀

templates/world-map-2d/packages/client/node/src/state-machine.ts
```diff
@@ -0,0 +1,98 @@
...
+      user!,
+      blockHeight,
+      {
+        input: "joinWorld",
+        ...parsedInput,
+      },
+      randomGenerator
+    )
+  );
+  for (let i = 0; i < result.length; i++) {
```
acedward: This for loop (referenced from the other issue)

templates/world-map-2d/packages/frontend/min.js
acedward, line 1: Not used?

templates/world-map-2d/packages/frontend/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "world-map-2d-frontend",
```
acedward: "name": "@world-map-2d/frontend",

Just for consistency

templates/world-map-2d/packages/shared/utils/src/mod.ts
```diff
@@ -0,0 +1,9 @@
+type VersionString = `${number}.${number}.${number}`;
+const VERSION_MAJOR = 1;
+const VERSION_MINOR = 1;
+const VERSION_PATCH = 1;
+export const gameBackendVersion: VersionString = `${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}`;
```
acedward: Ideally this is not used, we now have the verison in the `node/src/main.ts`

templates/world-map-2d/packages/shared/game-logic/src/mod.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { MatchState, MatchEnvironment, TickEvent } from './types';
+import { processTick } from './tick';
+
+export * from './types';
+export * from './tick';
+
+// We initialize the round executor object using lobby data + submitted moves + randomness generator.
+// This function extracts the match environment and match state from the lobby.
+// and the openworld `processTick` function
+export function initRoundExecutor(
```
acedward: If this is not used, it would be great.
These are legacy structures from v1, but it not possible it's OK

templates/world-map-2d/packages/shared/contracts/evm/solidity/README.md
```diff
@@ -0,0 +1 @@
+Add any custom Solidity contract for your project to this folder
```
acedward: I think this is not correct?
solidity go in /src/contracts/*.sol

templates/world-map-2d/packages/client/node/src/state-machine/v1/transition.ts
Lambdanaut: 👍

templates/world-map-2d/packages/client/node/deno.json
Lambdanaut: ah yeah. Those were copied over from another template. Removing them 👍

templates/world-map-2d/packages/client/node/src/api.ts
Lambdanaut: got it. Yeah, databases can be a pain in the butt with this requirement.

templates/world-map-2d/packages/client/node/src/state-machine.ts
Lambdanaut: 👍 got it

templates/world-map-2d/packages/frontend/min.js
Lambdanaut: You're right. It's not being used anymore. Not sure where it came from though. 
Removing it.

templates/world-map-2d/packages/shared/contracts/evm/solidity/README.md
Lambdanaut: 👍

templates/world-map-2d/packages/shared/game-logic/src/mod.ts
Lambdanaut: 👍

templates/world-map-2d/packages/shared/utils/src/mod.ts
Lambdanaut: 👍

templates/world-map-2d/packages/frontend/package.json
Lambdanaut: 👍

templates/world-map-2d/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+# Frontend / Explorer / Hardhat 1&2 / API / Batcher
+EXPOSE 8080
+EXPOSE 8545
+EXPOSE 9999
+EXPOSE 3334
+
+ENV RUN_IN_DOCKER=true
+ENV EFFECTSTREAM_STDOUT=true
+CMD ["/app/launch.sh"]
```
acedward, line 67: I guess this dockerfile doesn't work, this file is missing?

Lambdanaut: fixed 👍

PR #584, opened 2025-12-23
MQTT Batcher Server

Yiakman:
- Adds new MQTT server that updates based on input state
- New inputId field on BatcherInput that can be supplied by the client or generated by the batcher
- Unit tests to ensure MQTT updates specific input status


acedward:
mqtt migrated

PR #600, opened 2026-01-19
Feature/update btc midnight ledger6 lace

Yiakman:
This PR upgrades minor version to 0.5.0
Implements `midnight-balancing-adapter` that receives serialized unproven and unbalanced transaction to be finished and send by the batcher's own wallet.
Examples available in e2e midnight tests and e2e-wallets.
Additional features:
- Updates Night Bitcoin Template with Lace Wallet use.
- Minor fixes for Deno runtime verification
- Updates e2e-wallets to support lace, added counter contract interaction with and without balance delegation to the batcher with counter contract value display

remember in order to test e2e-wallets cd to the e2e/e2e-wallets dir after contract deployment and `deno task serve`


e2e/client/batcher/config.ts
```diff
@@ -2,12 +2,97 @@ import {
...
+};
+
+export const midnightBalancingAdapter = new MidnightBalancingAdapter(
+  DEFAULT_WALLET_SEED,
+  balancingAdapterConfig
+);
+
+
+// Bitcoin Adapter
+const BITCOIN_SEED = "my-super-secret-regtest-demo-seed-e2e";
```
acedward, line 88: In the config file we don't add specific adapter data. each goes in it's own file 🙏

e2e/client/batcher/main.ts
```diff
@@ -35,7 +35,7 @@ import { config, storage } from "./config.ts";
 
 import { bitcoinAdapter } from "./adapter-bitcoin.ts";
 import { effectstreaml2Adapter } from "./adapter-effectstreaml2.ts";
-import { midnightAdapter } from "./adapter-midnight.ts";
+import { midnightAdapter, midnightBalancingAdapter } from "./config.ts";
```
acedward, line 38: please import from adapter-midnight
and adapter-midnight-balancer 
so we can use these files in other projects as reference

PR #624, opened 2026-02-20
Celestia Sync Protocol + Primitives for blobs

acedward:
# Celestia

- [x] Sync Protocol
- [x] Primitive
- [x] E2E test config
- [x] E2E write blob
- [x] E2E read state-machine
- [ ] Local Binaries
- [ ] Create Launcher
- [ ] Add launcher at E2E   

NOTES: We are now reading though the RPC. Does this work with the light node as well?

### DIFF
--- a/e2e/client/node/e2e-tests/e2e.celestia.test.ts
+++ b/e2e/client/node/e2e-tests/e2e.celestia.test.ts
@@ -9,7 +9,7 @@ const CELESTIA_FEE = ENV.getNumber("CELESTIA_FEE", 2000);
 const CELESTIA_GAS_LIMIT = ENV.getNumber("CELESTIA_GAS_LIMIT", 100000);
 const CELESTIA_NODE_URL = ENV.getString("CELESTIA_NODE_URL", "http://localhost:26658");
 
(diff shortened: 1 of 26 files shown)


renaynay:
> NOTES: We are now reading though the RPC. Does this work with the light node as well?

For the first iteration, it's recommended for best stability / lowest latency retrieval to use RPC @acedward

acedward:
@renaynay thanks!

For developing/testing we are launching a local node and bridge.
For production environments we have the option of launching a light node or using some external RPC.
Even in this case you suggest going with an RPC is better?
NOTE: I didn't have luck with the links of production-quality providers in the documentation.

renaynay:
@acedward np!

Yes, I strongly recommend going with an RPC for the first iteration for the lowest possible blob retrieval latency. 

Do you mean these links didn't work for you? https://docs.celestia.org/operate/networks/mainnet-beta/#production-rpc-endpoints or rather that you're not satisfied with the offering of the providers listed there? Regardless, we have options for you.

I will contact you directly.

renaynay:
@acedward I reached out in the telegram groupchat.

PR #628, opened 2026-02-26
Support File Persistence with PGLite

yHSJ:
This change will allow users to use an on-disk file to persist PGLite state, configured via an environment variable. The default value is `memory://`, which means this change is backwards compatible.


acedward:
❤️

PR #631, opened 2026-03-02
Support all Provided NetworkIds

yHSJ:
In d5bdc7c, `MidnightNetworkConfig` changed to only support the strings `preview` and `undeployed`. Midnight actually provides the following `NetworkId` values:

```typescript
export declare const NetworkId: {
    readonly MainNet: "mainnet";
    readonly TestNet: "testnet";
    readonly DevNet: "devnet";
    readonly QaNet: "qanet";
    readonly Undeployed: "undeployed";
    readonly Preview: "preview";
    readonly PreProd: "preprod";
};
```

This PR supports using each of these networkIds, with a template config that follows the standards in Midnight. That being said, it is quite possible that not all of these values have running networks that are accessible in the same way as `preview` and `preprod`

### DIFF
--- a/packages/chains/midnight-contracts/src/midnight-env.ts
+++ b/packages/chains/midnight-contracts/src/midnight-env.ts
@@ -17,7 +17,7 @@ const env = (key: string | string[], fallback?: string): string => {
     "";
   }
   throw new Error('Invalid key type');
-}  
+}
 
 type NetworkConfig = {
     indexer: string;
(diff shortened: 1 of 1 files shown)


acedward:
Thanks. I wasn't sure the URLs would be standardized as `https://indexer.${networkId}.midnight.network/api/v3/graphql` 
but it looks is works for now :)

yHSJ:
I'm not 100% certain that's the case for other testnets, and what is even supported by Midnight now. (QaNet, DevNet, etc.)

That being said, I think it makes sense to *assume* that is the right configuration for them and, if they're not supported, that's on the Midnight team not on effectstream to handle. Though I'd totally understand another opinion and ultimately it's not my project.

PR #632, opened 2026-03-04
Midnight Nullifier Sync/Primitive

acedward:


packages/node-sdk/sync/src/sync-protocols/midnight/fetcher.ts
```diff
@@ -65,8 +66,18 @@ export class MidnightFetcher extends BaseDataFetcher<
...
       }`,
     );
+    const blockFetchOptions: BlockFetchOptions = {
+      contractActions: this.config.primitives.some(
+        (p) => p.primitive.type !== "Midnight:Nullifier",
+      ),
+      zswapLedgerEvents: this.config.primitives.some(
+        (p) => p.primitive.type === "Midnight:Nullifier",
+      ),
+    };
```
SebastienGllmt, line 90: probably you should also filter to check that it's for the same network as this sync protocol

acedward: The check was added when the MidnightClient is created, it's related primitives, must have the same network id.

PR #672, opened 2026-05-04
Feature/skip block hash storage

Yiakman:
Implemented a skipBlockHashStorage flag to avoid storing effectstream_block_hash values and save DB space.
In the case of Prando RNG seed we still compute the hash in memory exactly as before. Only DB write is ommited for the hash value on effectstream_blocks


acedward:
Implemented in another PR 👍

PR #675, opened 2026-05-07
fix: SyncChains events for skipped blocks & 0x0 empty block hashes

acedward:
## Summary
- **Fetcher optimization fix**: All fetchers (EVM, UTXORPC, Avail, Bitcoin, Celestia, Midnight) now always include the last block in their output even when it has no primitives. This ensures `SyncChains` events fire for every chain on every effectstream block, fixing the batcher timeout and e2e test hangs caused by the `fetchAllBlocks` optimization.
- **Batcher confirmation**: Changed `>` to `>=` in the batcher's `wait-effectstream-processed` logic — seeing the transaction's own block is sufficient since events are emitted after DB commit.
- **Empty block hashing**: Blocks with no operations (no migrations, no primitives, no scheduled data) store `0x0` as their hash. Previous `0x0` rows are deleted so only the latest empty block is kept. The in-memory hash chain only advances on real hashes, making restarts deterministic.

## Test plan
- [ ] Run e2e tests (previously stuck at batcher timeout and "schedule" game input)
- [ ] Verify `SyncChains` events fire for parallel chains even when blocks have no primitives
- [ ] Verify empty blocks store `0x0` hash and previous `0x0` rows are cleaned up
- [ ] Verify restart produces identical block hashes (deterministic hash chain)

### DIFF
--- a/packages/node-sdk/runtime/src/process-blocks.ts
+++ b/packages/node-sdk/runtime/src/process-blocks.ts
@@ -10,6 +10,7 @@ import type {
 } from "@effectstream/coroutine";
 import {
   blockHeightDone,
+  deleteEmptyBlocks,
   deleteScheduled,
   getFutureGameInputByBlockHeight,
   getFutureGameInputByMaxTimestamp,
@@ -133,6 +134,7 @@ export function* processFinalizedBlock(
     value,
     previousBlockHash,
   );
+  let hasOperations = false;
   try {
     /* STEP 0: Start the transaction. */
     yield* until(dbConn.query("BEGIN"));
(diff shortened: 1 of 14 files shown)


packages/node-sdk/runtime/src/process-blocks.ts
```diff
@@ -133,6 +134,7 @@ export function* processFinalizedBlock(
     value,
     previousBlockHash,
   );
+  let hasOperations = false;
```
acedward, line 137: generatePaimaBlockHash(...) should take into account other operations in this BH into account.

PR #681, opened 2026-05-11
feat(sync): enable has_certificate server-side filtering

acedward:
## Summary

- Enables server-side `has_certificate` filtering in the UTxO RPC Watch stream, removing the previous client-side fallback
- Adds `UtxorpcCertificatePattern` config schema supporting `stake_delegation` and `any_pool_keyhash` filter variants
- Pool delegation primitive now sends `any_pool_keyhash` when configured with a single pool, so Dolos only streams transactions involving that pool
- Tested end-to-end against the custom Dolos binary: delegations to the filtered pool are indexed, delegations to other pools are correctly excluded

> **Dependency:** merge after https://github.com/txpipe/dolos/pull/994 is merged into Dolos and a release is published.

## Test plan

- [x] E2E: delegate to filtered pool → indexed in ~10s
- [x] E2E: re-delegate to unfiltered pool → not indexed (server-side filtered out)
- [x] E2E: empty `has_certificate: {}` (match-any) → all 10 tests pass
- [x] Verified via `grpcurl` that Dolos returns apply events only for matching certificates
- [ ] Run full E2E suite after Dolos release with `has_certificate` support

### Note
Conway-era `stakeRegDelegCert` is not matched by `stake_delegation` — use `any_pool_keyhash` for cross-era compatibility.


acedward:
Waiting for next version after 1.1.1 that include watch certificates 
https://github.com/txpipe/dolos

PR #709, opened 2026-05-26
Update Midnight Bitcoin Template and Midnight package version updates

Yiakman:
Includes
- fixes for dust sync in the batcher
- fix bitcoin concurrency bug in the batcher that caused double spending
- updated template to use native unshielded token instead of custom erc20. Uses new UnshieldedSpend primitives to detect unshielded token transaction
- updated template to use the balancing adapter to pay dust fees instead of relying on NIGHT faucet to generate dust in the browser wallet.
- updates evm-midnight-v2 dependencies version updates


templates/run-template-tests.ts
```diff
@@ -27,7 +27,6 @@ const ENABLED = [
   // "evm-midnight",    // TODO: migrate to effectstream-bun
   // "minimal",         // TODO: migrate to effectstream-bun
   // "multi-chain-token-transfer", // TODO: migrate to effectstream-bun
-  // "night-bitcoin",   // TODO: migrate to effectstream-bun
```
acedward, line 52: we should add night-bitcoin-v2 into this list - ideally with some test; check the other templates 🙏

templates/evm-midnight-v2/packages/frontend/package.json
```diff
@@ -17,7 +17,7 @@
     "@midnight-ntwrk/compact-runtime": "0.15.0",
     "@midnight-ntwrk/compact-js": "2.5.0",
     "@midnight-ntwrk/dapp-connector-api": "4.0.1",
-    "@midnight-ntwrk/ledger-v8": "8.0.3",
+    "@midnight-ntwrk/ledger-v8": "8.1.0",
```
acedward: https://github.com/midnightntwrk/midnight-sdk/blob/main/COMPATIBILITY.md
do we have some reason to not follow the compat matrix?

acedward (changes requested):
FYI the test look ok, I think only hooking them in is missing.

templates/night-bitcoin-v2/start.dev.ts
```diff
@@ -0,0 +1,137 @@
...
+    dependsOn: ["create-wallets-bitcoin", "create-wallets-midnight", "mint-wallets-midnight"],
+  };
+});
+
+export default {
+  processes: [
+    ...launchPglite(),
+    ...launchBitcoin("@night-bitcoin/contracts-bitcoin", { resolveFrom: root }),
+    ...launchMidnight("@night-bitcoin/contracts-midnight", { resolveFrom: root }, {
+      env: { MIDNIGHT_STORAGE_PASSWORD: "YourPasswordMy1!" },
```
acedward, line 59: This will not work without link.sh , we use `cwd` now

acedward: to test ; cd template/night-bitcoin ; bun i ; bun dev should work

PR #713, opened 2026-05-27
Make pg_ivm optional via strategy pattern

Yiakman:
The engine hard-required the `pg_ivm` extension, blocking deployment on managed Postgres (RDS, Cloud SQL, Neon). This PR makes it auto-detected with a plain-SQL-view fallback gated behind an explicit opt-in, so nobody silently boots production on the degraded path.

### Changes

- `MaterializedViewStrategy` interface with `pgIvmStrategy` (emits `pgivm.create_immv`) and `plainViewStrategy` (emits `CREATE OR REPLACE VIEW`).
- `detectCapabilities()` probes the extension at startup via a soft-fail `CREATE EXTENSION` + `pg_extension` lookup.
- `selectViewStrategy(capabilities)` encodes the policy:
  - pg_ivm available → uses it silently
  - pg_ivm missing + `ALLOW_NO_PG_IVM=true` → returns `plainViewStrategy` and emits a prominent multi-line `WARN`
  - pg_ivm missing + flag unset → **aborts startup** with a remediation error (install link + opt-in instructions)
- Strategy threaded through `Primitive.dynamicTables`, all 10 IVM emitters, `createDynamicTables`, runtime startup, and the db-emulator.
- Dropped unconditional `CREATE EXTENSION pg_ivm` from the v-0.0.0 system migration.
- New "Required Extension: `pg_ivm`" section in `docs/site/docs/home/1000-effectstream-engine/1002-database.md` covering install, the opt-out, and the performance tradeoff.

### Backward compatibility

`pgIvmStrategy` keeps the legacy unsuffixed migration name (`dynamic-tables-<type>-<name>`), so existing deployments find their applied migration rows and skip re-application on upgrade. Only the new fallback strategy uses a `-view` suffix.

### Atomicity & performance notes

- pg_ivm v1.11 source verified: `IVM_immediate_maintenance` triggers run synchronously in the same transaction as the underlying write. The plain-VIEW fallback inherits this atomicity unchanged via the `AFTER ROW` trigger on `primitive_accounting`.
- The fallback is correct but read-bound: the intermediate tables grow monotonically (zero-balance rows aren't purged), so plain-view reads become seq scan + filter at scale. The fallback is **dev/test and low-volume only** — documented prominently and enforced via the opt-in gate.

### DIFF
--- a/docs/site/docs/home/1000-effectstream-engine/1002-database.md
+++ b/docs/site/docs/home/1000-effectstream-engine/1002-database.md
@@ -56,6 +56,40 @@ bun run --cwd packages/database pgtyped:update
 ```
 This introspects your SQL files and database schema, then writes a `*.queries.ts` file next to each `*.sql` with fully-typed wrapper functions.
 
+### Required Extension: `pg_ivm`
+
+Effectstream relies on the [`pg_ivm`](https://github.com/sraoss/pg_ivm) PostgreSQL extension to maintain the read-side views over each Primitive's intermediate state (`primitives.<view>`). `pg_ivm` produces an *incrementally maintained materialized view* — a compact, up-to-date projection that the engine writes through synchronously on each accounting update, so reads from `primitives.<view>` hit a small physical table instead of scanning the full intermediate.
+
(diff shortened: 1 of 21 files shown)


acedward:
This looks OK. Can we get the perf test.
I think we will have bottlenecks before the db performance, but just to be sure.

PR #717, opened 2026-05-28
Resilient pg pool error handling for serverless Postgres

Yiakman:
### Problem

Running against Neon (or any managed pg with idle eviction) crashed the
node with an unhandled `ECONNRESET` during TLS reconnect. We had no
process-level safety net and no way to distinguish a cold-start blip from
a real outage.

### Changes (3 commits)

**1. Unhandled-rejection logger + transient pg predicate**
- `installUnhandledRejectionLogger(component, shouldSurvive?)` in
  `@effectstream/log`: single process handler, logs at WARN if the
  predicate matches, ERROR + `process.exit(1)` otherwise.
- `isTransientPgError()` in `@effectstream/db`: narrow predicate matching
  `ECONNRESET` / `EPIPE` (with pg-stack guard) and three known pg/TLS
  messages. Walks `cause` chains. Five unit tests.
- `pg-connection.ts` installs the logger on first `getConnection()`.

**2. `PoolErrorTracker`**
- Tracks consecutive transient failures with two thresholds:
  - **≥ `SUSTAINED_THRESHOLD_MS` (60s)** → WARN → ERROR + `sustained-failure` tag
  - **≥ `FATAL_THRESHOLD_MS` (5min)** → `fatal` tag + `process.exit(1)` for orchestrator restart
- Both env-overridable (`POOL_SUSTAINED_THRESHOLD_MS` / `POOL_FATAL_THRESHOLD_MS`).
- Resets on successful `connect` *and* successful `runPreparedQuery`
  (the latter closes a warm-connection-reuse gap that would have caused
  false fatal exits hours later).
- `poolErrors.state()` exposed for consumers.

**3. ECONNREFUSED + /health integration**
- `ECONNREFUSED` and `"Query read timeout"` added to the transient set.
- `GET /health` now returns `poolErrors.state()` and flips to **HTTP 503**
  when `sustained === true` — usable as a readiness probe.

### Verification
[pool-error-e2e.zip](https://github.com/user-attachments/files/28328892/pool-error-e2e.zip)

- Unit: `bun test ./packages/node-sdk/db/src/transient-pg-errors.test.ts` — 5/5.
- E2E (attached, not committed): `pool-error-e2e.zip` drives the full
  baseline → kill pg → escalate → fatal exit → respawn cycle in ~35s. README inside.

```
eduardosoto@Eduardos-MacBook-Pro paima-engine % bun scripts/test-pool-health-e2e.ts
auto-detected brew postgres: postgresql@18
[+0.0s] config: SUSTAINED=10000ms FATAL=30000ms
[+0.0s] target: http://localhost:9999/health
[+0.0s] no orchestrator detected; launching one
           DB_HOST=localhost DB_PORT=5432 DB_USER=eduardosoto DB_NAME=effectstream_test PGLITE=false
[+0.2s] dropped+recreated database effectstream_test
[+0.2s] applying migrations...
[+6.7s] starting orchestrator (background)...
[+67.0s] FAIL timeout waiting for: orchestrator boot
           last observation: UNREACHABLE (Unable to connect. Is the computer able to access the url?)
[+67.0s] TEST FAILED: timeout: orchestrator boot
[+67.0s] stopping orchestrator...
eduardosoto@Eduardos-MacBook-Pro paima-engine % bun scripts/test-pool-health-e2e.ts
auto-detected brew postgres: postgresql@18
[+0.0s] config: SUSTAINED=10000ms FATAL=30000ms
[+0.0s] target: http://localhost:9999/health
[+0.0s] node already listening on http://localhost:9999/health — using it
[+0.0s] PASS baseline healthy  →  HTTP 200 ok consecutive=0 sustainedFor=0.0s sustained=false
[+0.0s] $ brew services stop "postgresql@18" >/dev/null 2>&1 &
    BREW_PID=$!
    while kill -0 $BREW_PID 2>/dev/null; do
      pkill -KILL -f "postgresql@18/bin/postgres" 2>/dev/null || true
      sleep 0.2
    done
    pkill -KILL -f "postgresql@18/bin/postgres" 2>/dev/null || true
[+0.9s] verified: pg unreachable (psql exit 2)
[+0.9s] PASS counter incrementing  →  HTTP 200 ok consecutive=1 sustainedFor=0.5s sustained=false
[+10.5s] PASS sustained=true and HTTP 503  →  HTTP 503 db-unreachable consecutive=11 sustainedFor=10.1s sustained=true
[+31.0s] PASS node exited (fetch fails)  →  UNREACHABLE (Unable to connect. Is the computer able to access the url?)
[+31.0s] $ brew services restart postgresql@18
[+32.5s] phase 7: polling for supervised restart (30s, best-effort)...
[+62.7s] note: no recovery within 30s — supervisor may not be configured (this is not a failure)
[+62.7s] ALL REQUIRED PHASES PASSED
```

### DIFF
--- a/packages/effectstream-sdk/log/src/mod.ts
+++ b/packages/effectstream-sdk/log/src/mod.ts
@@ -11,6 +11,7 @@ export {
 } from "./const.ts";
 export { attachTransport } from "./tslog.ts";
 export { DefaultLogLevels } from "./tslog.ts";
+export { installUnhandledRejectionLogger } from "./process-handlers.ts";
 
 // re-exporting this
 // so that we don't need to re-import opentelemetry in every component
--- a/packages/effectstream-sdk/log/src/process-handlers.ts
+++ b/packages/effectstream-sdk/log/src/process-handlers.ts
@@ -1,0 +1,38 @@
+import { SeverityNumber } from "@opentelemetry/api-logs";
+import { log } from "./mod.ts";
+
+/**
+ * Install a process-wide `unhandledRejection` handler.
+ *
+ * By default the handler logs the error at ERROR and re-crashes the process
+ * via `process.exit(1)` — preserving the "let it crash" semantic that Node /
+ * Bun apply when no handler is registered. The point of installing this
+ * handler is *visibility*: without it, Bun crashes silently with no log line
+ * about what the rejection actually was.
+ *
+ * Pass `shouldSurvive` to opt specific errors out of the crash — useful for
+ * known-transient failures from external systems (e.g. a pg pool ECONNRESET
+ * from a serverless Postgres cold-start). Surviving errors are logged at
+ * WARN; non-surviving ones at ERROR before exit.
+ *
+ * Idempotent: only the first call in this module registers a listener.
+ */
+let unhandledRejectionLoggerInstalled = false;
+
+export function installUnhandledRejectionLogger(
+  componentName: string,
+  shouldSurvive?: (reason: unknown) => boolean,
+): void {
+  if (unhandledRejectionLoggerInstalled) return;
+  unhandledRejectionLoggerInstalled = true;
+  process.on("unhandledRejection", (reason: unknown) => {
+    const survive = shouldSurvive?.(reason) ?? false;
+    log.remote(
+      componentName,
+      ["unhandledRejection"],
+      survive ? SeverityNumber.WARN : SeverityNumber.ERROR,
+      (log) => log(reason),
+    );
+    if (!survive) process.exit(1);
+  });
+}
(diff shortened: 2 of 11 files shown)


acedward:
a) runPreparedQuery does not run all queries.
b) So if a query fails, the unhandled exception will be caught ; but the original thread will die if it's not implemented to safely reexecute

Yiakman:
Instead of relying just on runPreparedQuery:

- Added processFinalizedBlockWithRetry: re-acquires a fresh client and reprocesses the block from BEGIN on transient errors, with backoff. `poolErrors` still escalates to a fatal exit when sustained; non-transient errors still crash.
- Destroy the dead client on release so the pool can't reuse the broken socket.
- markHealthy() on COMMIT so the write path feeds the tracker too

acedward:
I have to review this again, but I'm not sure what would happen in the sync protocols, e.g., in a primitive that fails. They are not in the begin/end loop

PR #763, opened 2026-06-10
feat(zswap-da): root-known + unshielded-existence liveness primitives

acedward:
Closes the last two liveness gaps in the ZSwap offer validator (LIVENESS plan §7/§7.1): an offer proving against a **fabricated or aged-out Merkle root**, and an offer referencing a **never-created unshielded UTXO**. Both now reject at STM ingestion and at `/api/zswap/submit` (the fee gate). Builds on #761 (validation + spent-set liveness) and #762 (offer fixture).

## SDK (`@effectstream/sm`, `@effectstream/sync`)
- Two new Midnight sync primitives:
  - **`Midnight:UnshieldedCreate`** — mirror of `UnshieldedSpend` over `unshieldedCreatedOutputs` (selected on the `Transaction` interface, so **system txs** that mint UTXOs are covered).
  - **`Midnight:ZswapRoot`** — emits the block's coin-commitment tree root (last RegularTransaction's `zswapMerkleTreeRoot`).
- `MidnightClient.fetchBlock` gains `unshieldedCreatedOutputs` + `... on RegularTransaction { zswapMerkleTreeRoot }` (inline fragment); `fetcher` dispatches the two primitives. Both registered in `builtInPrimitivesMap`.

## Template
- **`002-liveness-sets.sql`**: `created_unshielded` (append-only, from genesis — old UTXOs stay spendable) + windowed `known_roots`. Queries `InsertCreatedUnshielded`/`IsUnshieldedCreated`/`UpsertKnownRoot`/`IsKnownRoot`/`PruneKnownRoots` (prune keeps the latest root) — verified end-to-end against in-memory PGlite.
- **Handlers** `midnight-unshielded-create` / `midnight-zswap-root` populate the sets; root pruned to `ROOT_WINDOW_SECONDS` (default 14d; set ~3600 on a current 1h-window network), deterministic by block time. `celestia-zswap` + submit add `UTXO_UNKNOWN` + `ROOT_UNKNOWN` after the spent-set checks.
- **`validator/extract-root.ts`**: reads a shielded input's merkle root from the serialized `ZswapInput` (the `ledger-v8` binding exposes no getter), pinned to `zswap-input[v2]`, anchored on the nullifier (cross-checked against the getter). **Fail-closed** (`ROOT_UNREADABLE`).

## How the extractor was verified (the money-at-risk part)
The layout was derived and confirmed against **live-chain ground truth**, not guessed: a freshly-proven offer's extracted input root equals the indexer's `zswapMerkleTreeRoot` **byte-for-byte** (`packages/tests/discover-root-layout.ts`). Because it's always-on + fail-closed, a mis-parse yields bytes matching no known root → **reject, never accept-bad**; and the **accept path** (root ∈ `known_roots`) is unit-tested so legitimate offers aren't bricked.

## Tests
`bun test packages/validator packages/database` → **39 pass / 0 fail** — extractor layout (synthetic + real fixture), `accept-when-root-known` / `reject-when-unknown` (`ROOT_UNKNOWN`), `UTXO_UNKNOWN`, migration 002 + all 5 IRs over PGlite, prune-keep-latest.

## Caveats for the reviewer
- **Coordinated release**: the SDK changes (`node-sdk/sm`, `node-sdk/sync`) and the template ship together — dev resolves the SDK via `link.sh`; a published version is needed for a template release.
- **Full orchestrator e2e** is provided as `packages/tests/liveness-e2e.ts` (asserts the sets populate from live sync + a fresh offer is accepted). In my checkout it was blocked by a dev-stack DB-visibility quirk in the `link.sh`'d WIP SDK that hides **baseline** tables (`offer_file`, `known_tokens`) from the node API too — i.e. not introduced by this change. Please run it in a known-good environment to confirm the live plumbing.
- Upstream ask still open: a `ZswapInput.merkleTreeRoot` getter would replace the byte extractor with a one-liner.


acedward:
**All four primitives + the full lifecycle are now proven live** (commits `ce009ecd` + `ee5c847c`).

**Startup test-token mints (`ee5c847c`):** a new dev-only `midnight-mint-test-tokens` orchestrator process calls the offer-files contract's mint circuits headlessly (midnight-js `callTx`) — two shielded colors + one unshielded color to the genesis wallet. The unshielded mint emits `unshieldedCreatedOutputs` on chain, closing the last untested primitive: **`created_unshielded` populates live via `Midnight:UnshieldedCreate`**.

**Full lifecycle, run green end-to-end** (`packages/tests/full-lifecycle-e2e.ts` — `FULL LIFECYCLE PASS`; `liveness-e2e.ts` — `ALL PASS`):
- mint → offer **gives minted color A / wants minted color B** → `/api/zswap/submit` 200 (crypto + spent-sets + **root-known** against primitive-fed `known_roots`) → batcher → **Celestia** → ingestion (re-validated) → indexed
- settle on Midnight → **nullifiers consumed** (`spent_nullifiers` → 18) → offer **ARCHIVED (CONSUMED)** → `known_roots` advancing (9→12)
- garbage → `400 BAD_ENCODING`

**Earlier root-cause for reviewers using `link.sh` + monorepo SDK source** (`ce009ecd`): the dev-stack stall was a dual-wasm-instance failure — the monorepo-resolved fetcher and the template's compiled contract loaded separate `@midnight-ntwrk` wasm copies, so the contract's `ledger()` failed `instanceof` ("expected instance of ChargedState"), silently retried forever, stalling the merge so no block/migration committed. Dev fix: dedupe `onchain-runtime-v3` + `ledger-v8` store entries (published releases are single-tree, unaffected). The fetcher now logs per-height read failures. Also fixed: the `compact list` CLI hang no longer wedges boot.

**Honest residuals:** `spent_unshielded` settle-emission remains uncovered — the wallet facade can't yet build a well-formed OPEN offer with unshielded inputs (mixed intents come out give-only → `NOT_A_SWAP`; pure unshielded intents fail `wellFormed` with `SIGNATURE_INVALID` "unshielded offer action validation error"). Both were rejected correctly pre-fee, which itself exercises the validator's unshielded legs live; the e2e reports this as an explicit skip until the SDK supports unshielded open intents. Mint nonces are time-based (re-minting the same `(domain_sep, nonce)` recreates the identical commitment → node error 103); shielded mint colors are contract-derived, read from the call result.

---

**SDK-level e2e coverage added (`3ca23882`):** the `e2e/midnight` suite now registers both new primitives and asserts on their output end-to-end (fetcher → STM → user tables). A new `triggerUnshieldedCreates()` faucet helper does an unshielded NIGHT self-transfer from the genesis wallet; assertions check `primitive_accounting`, shape-valid `midnight_unshielded_creates` rows, strict root form (66 lowercase hex, `0x73` first byte), and a **fidelity cross-check** that the stored root byte-equals the local indexer's `zswapMerkleTreeRoot` for the same transaction. Suite green: **17 passed / 0 failed** (was 12); creates count 2 == exactly the one unshielded tx's outputs (non-vacuous), stored root byte-identical to the indexer.

PR #770, opened 2026-06-12
feat(midnight): TokenMint primitive — map token ids to minting contracts

acedward:
## What

Adds a new built-in Midnight sync primitive, **`Midnight:TokenMint`**, that maps a token id ("color") back to the **contract that minted it** plus the **domain separator** — the mapping wallets cannot provide (they only surface the 64-hex token type, and the derivation `rawTokenType = H(domain_sep, contract_address)` is one-way without this registry).

### How it works

A contract call's ledger `Effects` record mints as `domain_sep → amount` maps (`shieldedMints` / `unshieldedMints`). The indexer doesn't expose effects via GraphQL, so the fetcher selects the transaction `raw` bytes + `transactionResult`, deserializes with **ledger-v8** (same approach as the existing nullifier event decoder), walks `intents → actions → ContractCall` transcripts, and derives the token type with ledger-v8's own `rawTokenType()`. One event per `(tx, call, domainSep, kind)`.

Correctness details:
- **Failed segments roll back on chain** — the decoder consults `transactionResult`: `FAILURE` emits nothing; `PARTIAL_SUCCESS` applies guaranteed transcripts + only successful fallible segments (intent map key = segment id). System transactions (no `transactionResult`) are skipped — native NIGHT issuance is not a contract mint.
- Only `ContractCall` carries transcripts (deploys/maintenance updates cannot mint).
- Amounts travel as **decimal strings** (u64 mints exceed `MAX_SAFE_INTEGER`).
- The decoder never throws (tolerant like `zswap-decoder.ts`) — an undecodable tx logs and contributes nothing, so it can't wedge the fetch loop.
- The mint **nonce is not part of token identity** (it only randomizes coin commitments, and is never public for shielded mints) — the registry key is `(token_type, kind)`; `kind` matters because shielded/unshielded share the derivation formula.

### Changes

**SDK** (`packages/node-sdk/`):
- `sm/primitives`: `PrimitiveTypeMidnightTokenMint` + `midnight-token-mint/` module + registry entry (mirrors UnshieldedCreate).
- `sync/midnight`: new `mint-decoder.ts`; `MidnightClient` gains a flag-gated `raw` + `... on RegularTransaction { transactionResult }` selection; fetcher gains the dispatch branch, the fetch flag, **and the `contractActions` exclusion** (without it, an addressless primitive falls into the contract-state else-branch and crashes on `contractAddress!`).
- `sync/test/mint-decoder.test.ts`: tolerance unit tests (garbage bytes / FAILURE / missing result → `[]`).

**E2E** (`e2e/`, workspace deps — proves the primitive pre-publish):
- `counter.compact` gains `mint_shielded` / `mint_unshielded` circuits (copied from zswap-da's offer-files contract; compiles clean — launcher recompiles + redeploys per run; `ledger` declarations untouched so existing contract-state assertions are unaffected).
- New `trigger-token-mints.ts` (ported from zswap-da's `mint-test-tokens.ts`): mints one shielded + one unshielded token via `callTx` and returns the **wallet-visible colors**.
- Config/grammar/STM handler/table: registry `midnight_token_mints` with `UNIQUE (token_type, kind)`; on conflict `total_minted` accumulates (mapping + first-mint provenance immutable).
- Four assertion layers in `run-tests.ts`:
  1. `primitive_accounting` has `Midnight-TokenMint`;
  2. registry has shape-valid shielded + unshielded rows;
  3. **derivation fidelity** — every stored `token_type` byte-equals `rawTokenType(domain_sep, contract_address)` recomputed via ledger-v8;
  4. **the use case** — the trigger's wallet-visible colors resolve through the registry to the counter contract, with matching domain sep and exact `total_minted`.

## Verification

- `bun run e2e/runner.ts midnight` — green including the 4 new TokenMint assertions and all pre-existing ones (contract-state, Nullifier, UnshieldedCreate, ZswapRoot, batcher).
- **Negative check**: suite run with `doTriggerTokenMints()` disabled — the TokenMint assertions fail while the rest stays green (proves they exercise the real pipeline).
- `bun test ./packages`: 144 pass / 2 fail — the 2 failures are **pre-existing on v-next** (verified by stash-baseline: clean v-next = 141 pass / same 2 fail; this branch adds 3 passing decoder tests).
- Compact compile of the extended counter contract: clean (5 circuits).

## Notes

- WASM cost: one `Transaction.deserialize` per regular tx per block, **only when the primitive is configured** (flag-gated; nothing changes for existing consumers).
- Template adoption (e.g. zswap-da auto-registering mint provenance for `known_tokens`) is a **follow-up after the next publish** — templates pin published `0.100.22`.
- `contract-counter`'s `.` package export points at a non-existent `_index.ts` (pre-existing); the trigger imports by path like `contract-counter-deploy.ts` does.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -215,6 +215,7 @@
         "@effectstream/runtime": "workspace:*",
         "@effectstream/sm": "workspace:*",
         "@effectstream/utils": "workspace:*",
+        "@midnight-ntwrk/ledger-v8": "8.0.3",
         "@sinclair/typebox": "^0.34.30",
         "effection": "^3.5.0",
         "pg": "^8.14.0",
@@ -456,7 +457,7 @@
     },
     "packages/batcher": {
       "name": "@effectstream/batcher-sdk",
-      "version": "0.100.18",
+      "version": "0.100.22",
       "dependencies": {
         "@effectstream/crypto": "workspace:*",
         "@effectstream/event-client": "workspace:*",
@@ -499,7 +500,7 @@
     },
     "packages/binaries/avail-light-client": {
       "name": "@effectstream/npm-avail-light-client",
-      "version": "0.100.18",
+      "version": "0.100.22",
       "bin": {
         "npm-avail-light-client": "index.js",
       },
@@ -510,7 +511,7 @@
     },
     "packages/binaries/avail-node": {
       "name": "@effectstream/npm-avail-node",
-      "version": "0.100.18",
+      "version": "0.100.22",
       "bin": {
         "npm-avail-node": "index.js",
       },
@@ -521,7 +522,7 @@
     },
     "packages/binaries/bitcoin-core": {
       "name": "@effectstream/bitcoin-core",
-      "version": "0.100.18",
+      "version": "0.100.22",
       "bin": {
         "bitcoin-core": "./index.js",
       },
@@ -531,7 +532,7 @@
     },
     "packages/binaries/celestia": {
       "name": "@effectstream/celestia",
-      "version": "0.100.18",
+      "version": "0.100.22",
       "bin": {
         "celestia": "./index.js",
       },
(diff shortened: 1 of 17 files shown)


acedward:
domain sep is required for the full reconstruction, we should add the block height where the token was minted so you can get the event with the domain sep for full trustless execution?

PR #812, opened 2026-07-28
fix(sync): harden the sync service against external failures

acedward:
## What

Hardens the sync service against **external** failures — unreachable RPCs and
dead streams. The internal failure paths were already solid (transient-pg retry,
block-accurate resume, backpressure); the external ones were uneven, and their
dominant failure mode was *hang*, not *error*, which the existing error handling
could not see.

Every fix here landed only after a test proved the bug first. Those repros are
kept as permanent regression guards.

> **Reorg detection was split out into #813**, stacked on this branch. It
> was by far the most complex piece (new migration, new table, incident
> reporting) and deserves its own review.

## Why

All three of these end identically in production: the merge blocks on one
chain's page, block production stops, and **nothing reports a problem** — the
database stays perfectly healthy, so `/health` kept answering
`200 {"status":"ok"}`.

| # | Issue | Before |
|---|---|---|
| 1 | No RPC timeout on Bitcoin / NEAR / Avail | `readData` hangs forever; `consecutiveErrors` stays 0 |
| 2 | utxorpc had no `pollingInterval` | fetch loop starves the entire event loop |
| 3 | `startAsync` unsupervised | clean stream end = silent permanent stall; a throw kills the node |
| 4 | `/health` reported only DB reachability | all of the above invisible |

## Notable findings

**Cardano was one config field away from freezing.** The utxorpc schema was the
only sync protocol not merging `PollingSyncProtocol`, so `pollingInterval` was
undeclared, untyped and undefaulted — while the fetch loop keys its pacing off
exactly that property name. It works today only because every config in the repo
passes it anyway as an undeclared extra field. Built strictly from the schema,
the node froze: 200k loop iterations in ~4s while a 500ms `setTimeout` never
fired once. Not a slow node — a dead one, with no HTTP server, no other chain's
fetch loop, and (for a streaming chain) none of the callbacks that would let it
escape.

**A producer that returns cleanly was the worst failure found.**
`WatchMultiplexer.runWatcher` logs "stream ended unexpectedly" and returns; the
spawned task then completed *normally*. The chain received nothing forever while
every health counter stayed clean.

**`lastSuccessfulFetchMs` cannot detect a hang.** A caught-up chain never calls
`readData`, so its last-successful-fetch time is indistinguishable from that of a
chain wedged inside a request that will never return. `/health` keys off a new
`lastPollAtMs` heartbeat instead — the loop still cycles in the first case and
does not in the second.

## Testing

```bash
bun run e2e/sync-repro/run-tests.ts --docker
```

26 tests green (11 runtime reproductions + 15 sync-service). A new lightweight
`Dockerfile.sync-repro` gives the suite its own network namespace, so it can run
while other suites hold the PGLite gateway / API ports. It deliberately does not
reuse `.github/Dockerfile` (foundry + compact + solc + Playwright) — this suite
starts no chains and uses in-process PGLite.

## New config

| Field / env | Default | Purpose |
|---|---|---|
| `requestTimeoutMs` (per protocol) | 15s | bounds one RPC request |
| `pollingInterval` on utxorpc | — | now declared like every other protocol |

## Behaviour changes to review

- **`/health` can now return 503** — for `db-unreachable`, `stalled`, or
  `starting`. A chain erroring while blocks still flow is `degraded` + 200, so an
  orchestrator does not kill a node that is recovering on its own. Anything
  probing `/health` for liveness should be checked against this.
- **`pollingInterval` is now required on utxorpc configs.** Every config in this
  repo already passes it, so this is a no-op in practice — but it fails loudly at
  validation instead of freezing at runtime.

## Not included (pre-existing, flagged separately)

- `migrations/assets.ts` is stale: the bundled copy of `system-up-v-0.0.0` still
  contains `CREATE EXTENSION IF NOT EXISTS pg_ivm`, removed from the source in
  5d12bbcc. Migrations load from the bundle, so fresh databases still run it.
- `db-emulator` imports `@effectstream/sm` without declaring it, so its tests
  cannot load.

Both verified present at `d665de80`, before any of this work.

## Test status

`bun test ./packages` has 4 failures, all pre-existing and unrelated: 2
pool-error-tracker tests that pass in isolation and fail on cross-test
interference, plus the db-emulator import error. Verified not mine by swapping in
the base version of the one db file this branch touches — they persist unchanged.

### DIFF
--- a/.github/Dockerfile.sync-repro
+++ b/.github/Dockerfile.sync-repro
@@ -1,0 +1,50 @@
+# ══════════════════════════════════════════════════════════════════════════════
+# sync-repro test image
+# ══════════════════════════════════════════════════════════════════════════════
+# A deliberately small image for the `sync-repro` e2e suite. Unlike
+# .github/Dockerfile (which builds foundry, the Midnight compact compiler, solc
+# and Playwright Chromium because it runs the whole e2e matrix), this suite
+# needs none of that: it starts no chains, uses no orchestrator, and its
+# database is in-process PGLite.
+#
+# The reason to containerise it at all is isolation. The reproduction harness
+# binds a PGLite gateway port and an API port per node; running in a container
+# gives those their own network namespace so the suite can run while other test
+# suites are using the same ports on the host.
+#
+# Build + run:
+#   bun run e2e/sync-repro/run-tests.ts --docker
+# or directly:
+#   docker build -t effectstream-sync-repro -f .github/Dockerfile.sync-repro .
+#   docker run --rm effectstream-sync-repro
+FROM oven/bun:1.3.10-debian
+
+WORKDIR /app
+
+# Node is needed by postinstall scripts during `bun install` (core-js and
+# friends shell out to `node`), even though nothing here runs under Node.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+      ca-certificates \
+      curl \
+    && curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/*
+
+# node_modules is excluded by .dockerignore, so this is the source tree only.
+COPY . .
+
+# Not --frozen-lockfile: `packages/effectstream-sdk/wallets/npm` is a generated
+# build output that is present in the lockfile but not in the source tree, so a
+# frozen install always reports drift. .github/Dockerfile does the same.
+RUN bun install
+
+# PGLite (WASM) is the default backend; the Postgres-only consistency tests
+# skip themselves unless PGLITE=false. MQTT binds fixed ports that the harness
+# never frees, so it stays off.
+ENV PGLITE=true
+ENV MQTT_BROKER=false
+ENV EFFECTSTREAM_STDOUT=true
+ENV POLKADOTJS_DISABLE_ESM_CJS_WARNING=1
+
+# --in-container stops run-tests.ts from recursing back into docker.
+CMD ["bun", "run", "e2e/sync-repro/run-tests.ts", "--in-container"]
--- a/.github/Dockerfile.sync-repro
+++ b/.github/Dockerfile.sync-repro
@@ -33,10 +33,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # node_modules is excluded by .dockerignore, so this is the source tree only.
 COPY . .
 
-# Not --frozen-lockfile: `packages/effectstream-sdk/wallets/npm` is a generated
-# build output that is present in the lockfile but not in the source tree, so a
-# frozen install always reports drift. .github/Dockerfile does the same.
-RUN bun install
+RUN bun install --frozen-lockfile
 
 # PGLite (WASM) is the default backend; the Postgres-only consistency tests
 # skip themselves unless PGLITE=false. MQTT binds fixed ports that the harness
--- a/e2e/sync-repro/run-tests.ts
+++ b/e2e/sync-repro/run-tests.ts
@@ -67,6 +67,9 @@ async function runGroups(): Promise<number> {
       cwd: repoRoot,
       stdout: "inherit",
       stderr: "inherit",
+      // Spread order is deliberate: an ambient PGLITE=false must win, so the
+      // Postgres-backed consistency tests can be selected from the environment.
+      // PGLITE=true is only the default when nothing is set.
       env: { PGLITE: "true", ...process.env },
     });
     const code = await proc.exited;
--- a/e2e/sync-repro/run-tests.ts
+++ b/e2e/sync-repro/run-tests.ts
@@ -1,28 +1,46 @@
 /**
  * E2E suite: deterministic sync-service reproductions.
  *
- * Unlike the other suites, this one needs no orchestrator/chains. Each node runs
- * in its own subprocess over the synthetic `test` chain (so a restart is a
- * genuine new OS process whose in-memory Deque is gone while the committed
- * database survives — the production restart condition).
+ * Unlike the other suites, this one needs no orchestrator/chains. Two groups:
+ *
+ * 1. **Runtime reproductions** (`runtime/test/reproduction/`) — each node runs in
+ *    its own subprocess over the synthetic `test` chain (so a restart is a
+ *    genuine new OS process whose in-memory Deque is gone while the committed
+ *    database survives — the production restart condition).
+ *
+ * 2. **Sync-service reproductions** (`sync/test/`) — external-failure repros that
+ *    need no database: RPC clients pointed at a blackholed socket, and the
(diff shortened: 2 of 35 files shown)


acedward:
Thanks — all four items addressed, plus the nits. Three commits, all verified in the Docker-isolated suite (26 tests green).

## 1. `requestTimeoutMs` silently ignored on half the chains — fixed (`93664382`)

You were right that this was the one to fix rather than document, so I wired the rest instead of adding caveats:

- **Midnight** — the client already took the parameter; it just wasn't passed.
- **Celestia** — was pinned to `CELESTIA_RPC_TIMEOUT_MS`. That env var is now the *fallback*, so existing deployments keep their value. Because this client retries internally, the config bounds each **attempt** rather than the whole call — documented, since the semantics genuinely differ from the others.
- **EVM** — plumbed into the viem http transport, which is what the standing `TODO: transport options should come from the sync protocol configuration` in `createViemPublicClient` was asking for. `timeout` is a transport option, so it's destructured out of the override rather than spread onto the client config.

Only **utxorpc** remains unwired: it reads from a gRPC stream, not request/response RPCs, so the field is genuinely inert there — that path is covered by producer supervision instead. NTP and the synthetic chain make no network calls. Both exceptions are now in the README rather than left to be discovered.

**One extra find while auditing coverage:** `AvailClient.getLatestBlockHeight` did `const api = (await this.sdk).client.api;` and then never used `api` — only the commented-out `getHeader()` call ever needed it. That resolved the polkadot **websocket** SDK on the sync path, so if the node's websocket never connected, the call never returned and the chain stalled silently — precisely the failure this PR exists to prevent, sitting on the same code path I claimed to have bounded. `getBlockFromHash` is the only real SDK consumer and isn't on the sync path, so removing the dead `await` makes all of Avail's sync path bounded.

## 2. Backoff never de-escalated — fixed (`7681631c`)

Resets `attempt` when the producer survived longer than the max backoff, i.e. when it demonstrably worked. Also removes the cosmetic `2 ** attempt → Infinity`.

## 3. `consecutiveErrors` shared between loops — fixed (`7681631c`)

Producer failures now have their own `producerErrors` / `lastProducerErrorMs`. `/health` derives `erroring` from a failure inside a 60s window (2× max backoff), so a producer that recovers on its first or second restart clears the flag itself rather than waiting for the next block. The supervision test asserts the separation in both directions: a throw moves the producer counters and leaves `consecutiveErrors` at 0, and a clean return moves only the restart counter because a return isn't an error.

## 4. Unsafe cast in `health.ts` — fixed (`81921a6b`)

`startSync` now publishes the resolved interval as `SyncState.pollingIntervalMs` and both sides read it. Your point about the failure mode was the persuasive part: a silent fall-through to the 1s default would have *loosened* the wedged threshold, i.e. a monitoring bug that reports healthy.

## Nits (`81921a6b`)

- `sinceLastAppliedMs` narrowed to `number` in both the type and the schema.
- The `wedged`-during-catch-up caveat is documented where the threshold is defined, with the guidance to alert on the node's `status` rather than a single protocol's.
- Documented *why* `maxBufferedPages` has no default while its new neighbour does (the cap derives from `stepSize`, so a static default would override per-chain sizing) — agreed the mixed styles side by side were a copy-paste trap.
- Commented the deliberate `env` spread order in `run-tests.ts`.
- Left the NodeSource pipe unpinned per your read that it's acceptable for a test-only image.

Also added a **Producer supervision** section to the README — the supervision behaviour was user-visible but only documented in CLAUDE.md, and I'd referenced it from the timeout section.

#813 has been rebased onto this and is green (28 tests).

On your closing point: agreed that `/health` returning 503 is the thing to coordinate externally. Worth flagging that `starting` also 503s, so anything probing during the window between process start and the first applied block will see it — that's deliberate ("not ready to serve") but it's the case most likely to surprise an existing probe.

acedward:
## Determinism audit — one real bug found and fixed

Asked whether the merge is still deterministic after this PR. I audited each change rather than assuming, and **found one case where it wasn't**. Fixed in `af4f0a23`, with the property now pinned by a test in `38fb8379`.

### The bug: producer supervision could duplicate data (utxorpc only)

Supervising `startAsync` made restarts possible, and `WatchMultiplexer` was never written for that. All its state is on the instance, and its gRPC streams aren't cancellable from our side — so when **one** watcher's stream died, `Promise.all` rejected and aborted `start()` **while its siblings kept streaming**. The supervisor then called `start()` again, creating a second live stream per predicate feeding the same `pendingBlocks`, pushing every matched tx twice.

Why that breaks determinism rather than just leaking: duplicate txs → duplicate primitives → duplicate **scheduled STF inputs**. `primitive_accounting` has a unique index that absorbs duplicates, but scheduled inputs do not, so the STF would run twice on one node and once on another — divergence decided by which machine happened to lose a stream.

Exposure was bounded: `lowestIncompleteHeight` is monotonic, so completed heights could never be re-emitted. The window is the blocks in flight when the restart happened.

Fixed with a generation counter — a superseded watcher returns instead of handling its event (which also closes the stream via the async iterator's return path), and pending state at/above `lowestIncompleteHeight` is dropped on restart so the fresh stream re-delivers that window cleanly.

**This was introduced by this PR.** Before it, a throwing producer killed the node and a returning one silently stalled — both bad, neither corrupting.

### The other three changes are safe, and here's why

- **Timeouts** — a timeout makes `readData` throw, and `lastPage` only advances on a *complete* `DataFetched`, so the range is re-fetched rather than skipped. A slower machine that times out where a faster one doesn't converges on the same data. This would *not* hold if a partial page-range were ever committed, which is why the fetchers assemble the whole range before returning.
- **Pacing** — content per block is a pure function of height. The main clock derives timestamps arithmetically, and parallel data merges into the first root block whose timestamp is ≥ its own. Polling faster reaches the same answer sooner; wall clock only affects *how far* a chain has synced, never what block `n` contains.
- **`/health` + `lastPollAtMs`** — read-only observability, no merge input.

### Now pinned by a test

`determinism.test.ts` syncs two fresh databases to the same height under deliberately different timing — 5ms polling / stepSize 5 / 1s timeout versus 60ms / stepSize 100 / 30s timeout — and asserts identical content fingerprints. A second case isolates `effectstream_blocks` so a regression points straight at the hash chain. Both guard against passing vacuously on empty databases.

Suite: 26 → 28 tests, all green in Docker. #813 rebased and green (30).

### One harness bug found on the way

`runToHeight` silently dropped a custom `config` — only `runNode` forwarded it, though `node-runner` already preferred `spec.config`. The fallback to the default scenario meant tips no longer matched any protocol name, so the synthetic chain used its wall-clock fallback tip and synced to "now" (~28k blocks) instead of the target. Presents as a hang, so it's now forwarded and documented.

### Still worth knowing

The determinism guarantee is **per applied block**, not wall-clock reproducible: two nodes agree on the content and hash of block `n`, but not on *when* they reach it. That's the property consensus needs, and what the new test asserts. Coalescing also changes commit granularity (`sync_protocol_block_hash` in #813 records one row per commit), which is why that table is in the documented bookkeeping exclusions — flagged there rather than here.

PR #759, opened 2026-06-09
feat: add Solana chain support

mzf11125:
Closes #750

I am starting work on this with my AI agent. This PR implements full Solana support across all layers of the Effectstream framework following the decomposition outlined by @acedward in issue #750.

## What this PR covers

**A. Sync Protocol**

Adds `ConfigNetworkType.SOLANA` with a JSON RPC polling sync protocol backed by `@solana/web3.js`. The `SolanaClient` wraps `Connection` and handles skipped slots gracefully. The `SolanaFetcher` and `SolanaFetcherState` follow the same `BaseDataFetcher` and `SyncState` patterns used by EVM and NEAR. Block timestamps are monotonically non decreasing on Solana so equal timestamps are tie broken by slot number in `toRootPage`. A configurable `confirmationDepth` param controls finality delay vs security tradeoff.

**B. Primitives**

Adds three built in Solana primitives registered in `PrimitiveRegistry`:

- `SolanaAccountBalance` watches address balances per slot
- `SolanaProgramLogs` scrapes transaction logs for a program ID from `getBlock`
- `SolanaTokenAccount` watches SPL token account balances

**C. Batcher**

Adds `SolanaAdapter` in `packages/batcher/adapters/` with two modes:

1. Backend pays gas: batcher keypair signs as fee payer and submits directly
2. Dust sponsor mode: user partially signs client side and a `CapacityExchangeClient` calls a CES server to add the fee payer signature before submission, inspired by the SundaeSwap capacity exchange pattern at https://github.com/SundaeSwap-finance/capacity-exchange/blob/main/examples/react-vite/src/App.tsx#L128-L144

**D. Wallets**

Adds Solana wallet connector in `packages/effectstream-sdk/wallets/src/solana/` supporting Phantom, Backpack and Solflare via `@solana/wallet-adapter-*`. Also adds a `useSolanaCapacityExchangeWalletProvider` React hook in the frontend SDK. Ed25519 signature verification is added to the crypto package.

**E. Local Setup**

Adds `@effectstream/solana-node` binary package following the `@effectstream/bitcoin-core` pattern. Downloads `solana-test-validator` per platform at install time with a Docker fallback. An orchestrator launch script wires it into `bun run dev`.

**F. E2E Tests**

Adds `e2e/solana/` suite registered in `e2e/runner.ts` covering account balance sync, program log sync and batcher submission against a local validator.

**G. Template**

[...]

PR #815, opened 2026-07-29
feat(solana): Solana chain support (sync, primitives, gasless batcher, wallets, template)

### LINKED ISSUE #750: Midnight ↔ Solana Bridge: Add Solana chain support for Midnight-Solana cross-chain dApps
## Summary

Add **Solana** as a supported chain in Effectstream, enabling **Midnight-Solana** cross-chain dApps. This is distinct from the existing EVM-Midnight bridge. Solana brings a non-EVM, high-throughput L1 into Effectstream's chain abstraction layer, opening up private ZK state (Midnight) combined with Solana's DeFi ecosystem and SVM programmability.

## Motivation

Effectstream currently supports EVM, Midnight, Bitcoin, Cardano, Avail, Celestia, and NEAR. Solana is a major gap. It is the highest-throughput non-EVM L1 with a rich DeFi/NFT ecosystem. A Midnight-Solana bridge enables:

- **Private state on Midnight + Solana liquidity**. Deploy Compact contracts on Midnight with ZK proofs, settle or interact with Solana programs.
- **Cross-chain gaming**. Onchain games reading state from both SVM and Midnight in a single deterministic state machine.
- **Privacy-preserving DeFi**. Shielded token flows on Midnight bridged to Solana DEXs (Jupiter, Orca, etc.).
- **Sovereign rollups**. App-specific L2s that inherit finality from Solana with Midnight for private compute.

This is explicitly **not** an EVM bridge. It targets the Midnight-Solana pairing.

### Dust Sponsorship via 1AM Wallet

A key enabler for this bridge is the **1AM wallet's dust-free transaction flow**. On `preview`, `preprod`, and `mainnet` networks, the 1AM wallet's ProofStation sponsors all Midnight transaction fees. The bridge batcher needs zero NIGHT and zero DUST on all three live networks, which eliminates the DUST management headache entirely for staging, testing, and production deployments. For local `undeployed` development, the existing manual DUST registration flow still applies.

The 1AM ProofStation API exposes `POST /prove-and-balance` (combined proving + balancing, ideal for headless relayers), plus `POST /prove`, `POST /balance`, `POST /check`, `POST /verify`, `GET /wallet-status`, and `GET /health`.

## Proposed Work

Following Effectstream's existing chain integration patterns, this would involve:

### 1. Core SDK -- Chain Registration
- Add `ConfigNetworkType.SOLANA = "solana"` to `packages/effectstream-sdk/config/src/schema/network/types.ts`
- Add `ConfigSyncProtocolType.SOLANA_RPC_PARALLEL = "solana-rpc-parallel"` to sync protocol types
- Add Solana `ConfigNetworkSchema` to `packages/effectstream-sdk/config/src/schema/network/all.ts`
- Map `SyncProtocolToNetwork` for Solana
- Add Solana-specific primitive types (e.g. `PrimitiveTypeSolanaAccount`, `PrimitiveTypeSolanaProgram`)

### 2. Chain Package -- `@effectstream/solana-contracts`
- Create `packages/chains/solana-contracts/` following the pattern of `midnight-contracts/` and `evm-contracts/`
- Solana RPC client wrapper (using `@solana/kit` + `@solana-program/*`)
- Program/account subscription and event parsing
- Transaction building and submission utilities
- Local validator integration (Solana test validator for orchestrator)

### 3. Binary Wrapper -- `@effectstream/solana-validator`
- Create `packages/binaries/solana-validator/`. An NPM-wrapped Solana test validator binary.
- Follow the existing pattern of `@effectstream/midnight-node`, `@effectstream/avail-node`, etc.

### 4. Orchestrator Integration
- Add `launchSolana()` helper to generate orchestrator process config
- Wire into `orchestrator.config.ts` with feature flag (`DISABLE_SOLANA`)
- Process graph: `solana-test-validator` then `solana-validator-wait` then `deploy-solana-programs`

### 5. Sync Protocol
- Implement Solana RPC parallel sync in the node-sdk sync engine
- Subscribe to account changes via WebSocket (`accountSubscribe`, `logsSubscribe`)
- Parse Solana account data into Effectstream's internal event format
- Handle Solana's slot-based block model (distinct from EVM blocks and Midnight blocks)

### 6. Template -- `midnight-solana-v2`
- Create `templates/midnight-solana-v2/` modeled after `templates/evm-midnight-v2/`
- Example Compact contract on Midnight + Anchor program on Solana
- Shared state machine bridging both chains
- Batcher configuration for cross-chain transactions
- React frontend with Midnight wallet (1AM/Lace) and Solana wallet adapter

### 7. E2E Tests
- Create `e2e/solana/` test suite
- Infrastructure validation (validator launch, program deploy)
- Sync tests (assert Solana account state in Effectstream DB)
- Cross-chain tests (Midnight Compact to Solana program interaction)

## Technical Considerations

| Area | Detail |
|------|--------|
| **Block model** | Solana uses slots (not blocks). Slots are approx 400ms, produced optimistically, not finalized until 32 slots deep. Sync must handle fork resolution. |
| **Account model** | Solana accounts have owners, rent, and data. This is very different from EVM storage or Midnight ledger. Account data parsing needs schema definitions. |
| **Transaction model** | Solana transactions batch multiple instructions atomically, with compute budget and priority fees. Distinct from EVM or Midnight tx formats. |
| **Program deployment** | BPF bytecode deployed to accounts. Upgrades via buffer accounts. Orchestrator needs to handle this flow. |
| **Midnight side** | Compact contracts with `disclose()`, commitment/nullifier patterns, and shielded/unshielded token flows. Cross-chain state sync via Indexer GraphQL. |
| **Batcher** | Cross-chain tx batching needs a Solana adapter (like the existing EVM and Midnight batcher adapters). May need to handle Solana's different fee model. |
| **1AM Dust Sponsorship** | The 1AM wallet's ProofStation on preview, preprod, and mainnet eliminates DUST management for the bridge batcher. Key API route is `POST /prove-and-balance`. Zero NIGHT and DUST needed for all live network deployments. |

## Dependencies

- `@solana/kit` + `@solana/kit-plugin-rpc` + `@solana/kit-plugin-signer`
- `@solana-program/system` and `@solana-program/token` for instruction builders
- Solana CLI tools (`solana-test-validator`, `solana`, `anchor`)
- `@midnight-ntwrk/wallet-sdk-*` for Midnight wallet integration

## Related

- Existing EVM-Midnight bridge: `templates/evm-midnight-v2/`
- Chain integration patterns: `packages/chains/evm-contracts/`, `packages/chains/midnight-contracts/`
- Config schema: `packages/effectstream-sdk/config/src/schema/`
- Contribution guide: `docs/site/docs/home/1000-effectstream-engine/1100-contributions.md`

## Scope

This is a large feature. I'm proposing to contribute it incrementally in phases:
1. Core SDK registration + Solana network schema (smallest)
2. Binary wrapper + orchestrator integration (local dev loop)
3. Chain package + RPC client (sync + read)
4. Sync protocol implementation
5. Template + e2e tests
6. Batcher adapter

Feedback welcome on scope, priorities, and any architectural concerns.

acedward:
Closes #750 · supersedes #759

Solana as a first-class chain: config schema, sync protocol, two primitives, a gasless fee-payer-sponsor batcher, wallet + Ed25519 verification, an e2e suite, and a starter template.

Originated as **#759** by @mzf11125, moved in-repo so it actually gets CI — that PR is `on: [push]`-only from a fork, so its runs land in the fork's Actions tab and #759 itself shows *"no checks reported"*. It had never been validated by this repo.

## ✅ CI green on the current head

[Run 30501969885](https://github.com/effectstream/effectstream/actions/runs/30501969885) @ `5bdbb749` — `changes` ✓ 17s · `e2e` ✓ 27m56s · `frontend-build` ✓ 27s · `ci-ok` ✓ (`template-tests` skipped, see gap #1)

```
All 11 suite(s) passed        [PASS] solana (72.8s)
```

Green on two consecutive runs — [30500289340](https://github.com/effectstream/effectstream/actions/runs/30500289340) validated the Agave pin, this one adds the final fixes and docs.

The solana suite is a real end-to-end run: validator boot → airdrop → transfer → sync node picks up `solana-program-log` → batcher sponsors a gasless Memo tx → row lands in `solana_log_events` → a non-target-program instruction is correctly rejected.

## What changed after #759

The config/schema layer arrived in good shape and is essentially untouched. Everything below was found by review or by CI on this branch.

### Blocking bugs

| | Fix |
|---|---|
| `SolanaSyncState` never implemented `outputToLastPage` — declared `abstract` on `SyncState`, called unconditionally by the merge. The sync node threw `TypeError` on the first merged Solana datum, and block-accurate resume couldn't work | [`4c9a09c4`](../../commit/4c9a09c4) |
| No `bufferAtCap` gate in `stateToInput` — all nine other protocols have it first. Without it the fetch loop has no backpressure and buffers toward the whole backlog during catch-up (sync `CLAUDE.md` finding #1) | [`4c9a09c4`](../../commit/4c9a09c4) |
| **Agave 4.1.2 could never start in CI.** ≥3.1 hard-asserts io_uring on Linux; Docker's default seccomp blocks those syscalls and the whole e2e suite is containerized. macOS builds don't compile the assert in, so it passed every local check and died in CI. Bisected `4.1.2 ✗ · 4.0.3 ✗ · 3.1.14 ✗ · 3.0.14 ✓ · 2.3.13 ✓` and pinned 3.0.14 | [`8ba48577`](../../commit/8ba48577) |
| Entirely-skipped slot range never advanced the page, so `stateToInput` re-requested the same window forever — a permanent stall with no recovery once blocks resumed | [`5bdbb749`](../../commit/5bdbb749) |

### Security

| | Issue | Fix |
|---|---|---|
| **A1** | `validateInput` waved through any `ComputeBudget` instruction. The sponsor is the fee payer, so an uncapped `SetComputeUnitPrice` is user-controlled spend — one transaction could drain it, while the docblock claimed the sponsor "cannot be drained regardless of any operator policy" | `maxPriorityFeeMicroLamports`, default `0n`; unknown discriminators rejected rather than assumed harmless — [`cf482f06`](../../commit/cf482f06) |
| **A3** | `verifySignature` checked that *some* non-sponsor signature was valid but never tied it to `input.address`, so submissions could be attributed to a third party and rate-limited under an identity that never signed | Claimed address must be among the signers — [`cfbe79af`](../../commit/cfbe79af) |
| **B1** | **`SOLANA:ProgramLog` was spoofable.** It gated on `accountKeys.includes(programId)` plus a substring match over *every* log line. Naming a program as a bare account key doesn't invoke it, so a different program could `msg!` the event string and the primitive would report it as the watched program's event with an attacker-chosen payload | New `program-logs.ts` walks the `invoke`/`success` framing; invocation is now the authorization boundary. 10 unit tests incl. the attack — [`3c572a35`](../../commit/3c572a35) |
| **B2** | Reverted transactions were processed — `meta.err` captured but never checked, so rolled-back logs and pre-state balances drove state transitions | Skipped, plus `meta: null` no longer dereferenced — [`fde18c7d`](../../commit/fde18c7d) |
| **B3** | `SOLANA:AccountBalance` looked up only `message.accountKeys` (static keys), so an address pulled in via an **address lookup table** was silently invisible | `resolveAccountKeys` rebuilds the balance-indexed list — [`5692dc48`](../../commit/5692dc48) |
| **B4** | `toMsTimestamp` returned `0` for a null `blockTime`, sorting that block to the unix epoch and breaking the monotonicity the merge gate depends on | Resolved at fetch time with carry-forward; `Output.blockTime` narrowed to `number` — [`5692dc48`](../../commit/5692dc48) |
| **C1** | The validator binary was downloaded with no integrity check | SHA-256 of the extracted binary verified **before first execution** — [`60c873b5`](../../commit/60c873b5) |
| **C2** | Pinned to EOL `solana-labs` 1.18.26, and the `aarch64-unknown-linux-gnu` URL **404s** (upstream publishes no such asset) | Moved to `anza-xyz/agave`, dropped the phantom target — [`60c873b5`](../../commit/60c873b5) |
| **C3** | Two Ed25519 secret keys committed with no guard rail | Batcher refuses to start when the committed sponsor key is paired with a non-loopback RPC; gitignore + README warning — [`60c873b5`](../../commit/60c873b5) |
| **C4** | `--bind-address 0.0.0.0` exposed the unauthenticated RPC to the network | Defaults to `127.0.0.1` (`SOLANA_BIND_ADDRESS` to override). The **faucet** ignores this and always binds all interfaces — Agave offers no flag; documented — [`60c873b5`](../../commit/60c873b5) |

### Quality

- `SolanaProvider.signTransaction` was base58 while claiming to match the batcher's base64 contract — the adapter would have decoded garbage and rejected everything. Now base64 both sides; its test asserted the wrong encoding and now asserts base64 round-trips *and* that base58 doesn't.
- `root-page.test.ts` had a case that was literally `expect(f(x)).toBe(f(x))` — true of any implementation. Rewritten to assert real properties.
- Validator startup output is now captured (40-line rolling tail) instead of discarded unless `--verbose`. The validator reports startup failures on **stdout**, so CI failures previously surfaced as a bare `exited with code 1` — this is what made the io_uring bug findable at all.
- `e2e` job `timeout-minutes: 30 → 45`. Green runs already sat at 28–29 min, so one `E2E_MAX_RETRIES` retry blew the budget and the job was killed mid-retry, reported as `cancelled` — indistinguishable at a glance from "superseded by a newer push".
- Exported `SolanaAccountBalancePrimitive`, dropped unused `SolanaClient.getBlocks`, un-mangled a `NearPrimitive` comment.

## Docs

- New [`200-chains/211-solana.md`](docs/site/docs/home/200-chains/211-solana.md) — config, both primitives, the sponsor batcher **and its limits**, wallets, crypto, orchestration. Contracts renamed 211 → 212 so Solana sorts with the chains.
- Chains index bullet + feature-support matrix row; sync-service protocol table; primitives table; orchestrator launcher lists.
- New `packages/binaries/solana-node/README.md` (auto-synced to `540-binaries`) covering the integrity check, the loopback bind and faucet caveat, the platform matrix, and the Agave pin with its bisection.
- Solana added to the sync + batcher package READMEs, `CLAUDE.md`, and the template guidelines' chain/adapter/grammar tables.
- Docs site builds clean (only pre-existing blog `#whats-next` anchors).
- `e2e/shared/contracts/solana` ([`bf601fab`](../../commit/bf601fab)) is the reference deploy other templates can crib `launchSolana` from — this was a known gap and is now closed.

## Commits

Read in order, this is also the story of what CI could and couldn't tell us.

| | |
|---|---|
| `e98a0dcf`…`9949610e` | #759 as authored by @mzf11125 (8 commits), plus the earlier in-repo fixes |
| [`4ec41f17`](../../commit/4ec41f17) | merge `v-next` — the fork branch was 19 commits behind |
| [`4c9a09c4`](../../commit/4c9a09c4) | `outputToLastPage` + `bufferAtCap` — the two `SyncState` contract holes |
| [`cf482f06`](../../commit/cf482f06) | cap the priority fee the sponsor pays (A1) |
| [`fde18c7d`](../../commit/fde18c7d) | stop emitting primitives for reverted transactions (B2) |
| [`3c572a35`](../../commit/3c572a35) | attribute program logs to the emitting program (B1) |
| [`cfbe79af`](../../commit/cfbe79af) | require the claimed address to have signed (A3) |
| [`5692dc48`](../../commit/5692dc48) | resolve lookup-table addresses; drop the null-`blockTime` → 0 mapping (B3, B4) |
| [`60c873b5`](../../commit/60c873b5) | pin + verify the validator binary, loopback bind, guard the dev keys (C1–C4) |
| [`8c46e82d`](../../commit/8c46e82d) | correct the scope of the bin-wrapper arch bug |
| [`71e87d14`](../../commit/71e87d14) | surface validator startup output; raise the e2e timeout — **this is what made the next commit findable** |
| [`8ba48577`](../../commit/8ba48577) | pin Agave 3.0.14 — 3.1+ cannot start in Docker (io_uring) |
| [`38229c17`](../../commit/38229c17) | correct the io_uring note: seccomp *is* the blocker on a real kernel |
| [`5bdbb749`](../../commit/5bdbb749) | remaining known issues + full chain docs |
| [`eed32325`](../../commit/eed32325) | solana-starter: fix the fresh-clone break; 14/14 locally |

## Known gaps — deliberate, not overlooked

1. ~~**`solana-starter` not in `ENABLED`**~~ — **closed.** 0.102.0 shipped Solana support, so the template now installs from npm with no `link.sh` and is enabled in `templates/run-template-tests.ts` ([`da7ae814`](../../commit/da7ae814)). Running it standalone for the first time exposed three defects that would have broken it for anyone installing from npm — `resolveFrom` not seeing workspace packages, a 60s `chain:wait` shorter than the validator download, and the vendored `solana` CLI's airdrop returning 400 against its own test validator — all fixed. **14/14 pass** through the real CI harness.

2. **No `tsc --noEmit` job.** The `outputToLastPage` bug was a plain type error that no job would have caught — but the repo carries a **229-error** pre-existing baseline, so a blocking gate isn't viable without either fixing those or adding baseline-comparison machinery. Out of scope; worth its own issue.
3. **A2 — no rate limit.** The adapter bounds *per-transaction* cost but not *volume*; every accepted tx costs the sponsor the 5000-lamport base fee. `RateLimitStore` needs wiring in the template before a funded batcher is exposed publicly. Documented in the adapter docblock and README.
4. **Agave pinned to 3.0.x.** A CI-configuration constraint, not an upstream dead end: on a kernel supporting io_uring the block is purely seccomp (`default → EPERM`, `unconfined → OK`, verified natively). A minimal seccomp profile allowing `io_uring_setup/enter/register` for the e2e container unlocks 4.x. Worth filing upstream too — a hard `assert!` that panics in any container is arguably an Anza bug.
5. **Checksum verification exists only in `solana-node`.** The other 11 `packages/binaries/*` still download unverified; the pattern should spread.
6. **No SPL token-balance primitive.** This ships `SOLANA:ProgramLog` and `SOLANA:AccountBalance` only. A third built-in, `SolanaTokenAccount`, is named in #759's description; I went looking for it to carry across and could not find an implementation on any ref I have access to (details in the thread below — if I'm searching the wrong place, a commit pointer would sort it out immediately).

   Either way it's worth having as a follow-up. It needs `resolveAccountKeys` (the B3 treatment), since `meta.pre/postTokenBalances` carry an `accountIndex` into the same resolved list, plus the B2 reverted-transaction skip.
7. `SolanaPrimitive` is an undiscriminated union of optionals, forcing truthiness dispatch in the fetcher and an `as` cast in `getConfig`. `Output.transactions` is buffered per block but nothing consumes it. `readData` does one RPC round-trip per slot, serially.


mzf11125:
Thanks @acedward, I'll continue working on this

mzf11125:
Closed #759 in favor of this. Thanks for the writeup. The invoke-framing and lookup-table findings were things I had wrong at the root rather than at the edges, and the revert-and-confirm-it-fails checks on the new tests are a habit I'm taking from this.

Two notes on the current state:

**A primitive dropped between the two PRs.** #759 had a third built-in, `SolanaTokenAccount`, watching SPL token account balances. This ships ProgramLog and AccountBalance only, and it isn't in the known-gaps list, so I assume it fell out during the rework rather than being cut deliberately. Happy to add it in a follow-up if you want it. It would need the same `resolveAccountKeys` treatment as B3, since `meta.pre/postTokenBalances` carry an `accountIndex` into the same resolved list, plus the B2 revert skip.

**Known gap 6 is stale.** `bf601fa` added `e2e/shared/contracts/solana`, so the "no reference deploy" item is closed.

On the remaining gaps, I'd like to take these unless someone is already on them:

- Gap 3, wiring `RateLimitStore` in the template. Last real hole before a funded batcher can be exposed publicly, and it completes the A1 cap.
- Gap 4, a minimal seccomp profile allowing `io_uring_setup/enter/register` for the e2e container, which unpins Agave from 3.0.x.
- Gap 5, spreading the SHA-256 check to the other 11 `packages/binaries/*`.

Can also file your generic `link.sh` note from #759 as its own issue so it doesn't get lost when that PR closes.

acedward:
Thanks — and the revert-and-confirm-it-fails habit is worth the small cost every time. It caught two things here that would otherwise have shipped as green: the reverted-transaction skip had a test that passed with the guard removed until I fixed the fixture, and the e2e spoof case only started failing against the pre-fix fetcher once it gated on the same slot having synced.

Gaps 3, 4 and 5 aren't in flight on my side, so they can be addressed in a separate PR whenever it suits — they're independent of each other and of this branch. The generic `link.sh` note from #759 is worth its own issue too, so it doesn't disappear when that PR closes.

Gap 6 was stale; you're right. `bf601fab` added `e2e/shared/contracts/solana`, so I've closed that item and moved the SPL token primitive into its place in the body.

## On `SolanaTokenAccount` — I went looking and came up empty

I tried to carry it across and couldn't find an implementation to carry. Before I write it from scratch, can you point me at the commit? I'd rather port your version than reinvent it, and it's very possible I'm searching the wrong ref.

What I checked, against a fresh `git fetch origin refs/pull/759/head` (`9949610e`):

```
$ git ls-tree -r --name-only refs/remotes/pr759 -- packages/node-sdk/sm/primitives/src/ | grep -i solana
packages/node-sdk/sm/primitives/src/solana-account-balance/solana-account-balance-grammar.ts
packages/node-sdk/sm/primitives/src/solana-account-balance/solana-account-balance-primitive.ts
packages/node-sdk/sm/primitives/src/solana-program-log/solana-program-log-grammar.ts
packages/node-sdk/sm/primitives/src/solana-program-log/solana-program-log-primitive.test.ts
packages/node-sdk/sm/primitives/src/solana-program-log/solana-program-log-primitive.ts

$ git log --all --oneline -S "SolanaTokenAccount"          # (no output)
$ git log --all --oneline -S "PrimitiveTypeSolanaTokenAccount"   # (no output)
$ git log --all --oneline -S "postTokenBalances"           # (no output)
```

I also diffed the two trees directly rather than eyeballing: every Solana-touching path in #759 (91 files) is present here (now 110), and the set of `PrimitiveTypeSolana*` identifiers is identical across both — `AccountBalance` and `ProgramLog`, nothing else.

The description does list three, so my working assumption is that the third was planned and described but not landed. If it exists on a branch that never made it into #759's head, `-S` over `--all` wouldn't see it from here.

## One thing genuinely *was* lost, and it wasn't that

Your instinct was right, just aimed at the wrong casualty. When I merged `v-next` I resolved the `wallets/package.json` conflict with `git checkout --theirs` and re-added only `@solana/web3.js`. That takes the whole file, so it silently dropped everything else the Solana work had put there:

- exports: `./solana`, `./solana-local`
- dependencies: `@noble/curves`, `@wallet-standard/app`, `bs58`

Fixed in `55066414`. The missing exports were the dangerous half — `@effectstream/wallets/solana-local` would not have resolved for any consumer, and since nothing in this repo imports it, no test would have caught it. It surfaced only because CI's clean `bun install --frozen-lockfile` + browser build failed where my warm `node_modules` passed.

After fixing it I ran the same file-set and identifier comparison above across the whole merge to check for other casualties. That file was the only one.

## Notes for whoever picks up the token primitive

Your read on it matches mine — `resolveAccountKeys` for the `accountIndex` into the resolved list, plus the B2 revert skip. Both already exist, so it should be additive for whoever takes it:

- `resolveAccountKeys` is exported from `sync-protocols/solana/SolanaClient.ts`
- the revert skip is one line at the top of the per-tx loop in `readPrimitives`
- `read-primitives.test.ts` already has fixtures for the ALT and reverted-tx cases that a token primitive can reuse

Worth knowing before starting it: the config type declares `scheduledPrefix` while the runtime reads `stateMachinePrefix` (`runtime/src/main.ts` spreads the config straight into the `Primitive` constructor). Setting only `scheduledPrefix` still writes accounting rows but never reaches the state machine, silently. I added `stateMachinePrefix` to `BasePrimitive` and documented which one wins, but the NEAR docs and e2e config still use the other — NEAR's tests pass because they assert on `primitive_accounting`, which is written either way.

mzf11125:
Checked all of this against `v-next` and the branches. It holds up, including the part I would rather it did not.

## The token primitive was never written

I went looking too, and came up empty in a way that answers your question. I checked wider than 759's head, because the description was mine and I wanted to know whether I had written it somewhere and then lost it:

```
$ for b in feat/solana-support feat/solana-sync-primitives-batcher solana-clean feat/solana-starter-ci-enable
> do echo "== $b"
>    gh api "repos/effectstream/effectstream/git/trees/$b?recursive=1" --jq '.tree[].path' \
>      | grep "primitives/src/solana"
> done
```

All four branches return the same two directories, `solana-account-balance` and `solana-program-log`. `search/code` for `SolanaTokenAccount` across the repo returns `total_count: 0`, and `git log --all -S SolanaTokenAccount` over my fetched fork remote returns nothing either.

So your working assumption is the correct one. Three primitives were described, two were written. Task 4 on that checklist got ticked when the two landed and I never went back to split it. There is no commit to point you at, and no ref you were searching wrong. Sorry for sending you looking.

I will write it fresh rather than leave it named and absent. Your three pointers all check out where you said they are, so it should be additive:

* `resolveAccountKeys` is exported at `SolanaClient.ts:55`
* the revert skip is `fetcher.ts:278`, first line of the per transaction loop in `readPrimitives`
* `read-primitives.test.ts` has the ALT and reverted transaction fixtures

## On the prefix trap, I scoped how far it reaches

Only one config is actually misconfigured today. Every other suite already passes `stateMachinePrefix`:

| config | `scheduledPrefix` | `stateMachinePrefix` |
|---|---|---|
| `e2e/near/config.ts` | 6 | 0 |
| evm, cardano, midnight | 0 | 6 each |
| bitcoin, avail, celestia | 0 | 1 each |
| `e2e/solana/config.ts` | 0 | 3 |

The two template hits are false positives: `templates/preorder/packages/node/primitives.ts` and the `mct-erc1155` custom primitive return `scheduledPrefix` from `getConfig()`, which is the outbound direction and matches every builtin. They need nothing.

What makes NEAR worth fixing rather than just documenting: `e2e/near/node.ts` registers six state transitions, one per prefix, and `e2e/near/grammar.ts` maps all six. None of those six handlers has ever run. The suite was written expecting state machine delivery and asserts one layer below it, so it stayed green while the top half was dead. Your read on why is exactly right, and it is a stronger case than the docs mismatch alone.

That makes the fix six lines in `e2e/near/config.ts`, six in `210-near.md`, and one assertion above `primitive_accounting` so it cannot go quiet again. I will carry that with the gap work rather than leave it as a note.

## Merge casualty confirmed

`55066414` restores all six on `v-next`. `./solana` and `./solana-local` are back in `exports`, and `@noble/curves`, `@wallet-standard/app` and `bs58` are back in `dependencies`. The clean install reproduction is the useful part of that story: a warm `node_modules` hides a dropped export entry completely when nothing in the repo imports it.

## Taking these

Gaps 3, 4 and 5, plus the generic `link.sh` note as its own issue. Separate PRs since they are independent of each other, in that order, with the NEAR prefix fix folded into the first one.

PR #730, opened 2026-05-28
chore(templates/zswap-da): bump @effectstream/* to 0.100.18

acedward:
## Summary
- Pins every `@effectstream/*` dependency in `templates/zswap-da` to `0.100.18`.

## Test plan
- [ ] `LINK_LOCAL=1 bun run templates/run-template-tests.ts zswap-da` — FAILS locally (3m6s) due to a vendored `celestia-appd` requiring **GLIBC 2.38** which the local glibc lacks. Same failure mode at 0.100.17; **environmental, not a regression**. Expected to pass on hosts/CI runners with glibc ≥ 2.38.

PR #720, opened 2026-05-28
chore(templates/batcher-validations): bump @effectstream/* to 0.100.18

acedward:
## Summary
- Pins every `@effectstream/*` dependency in `templates/batcher-validations` (root + 5 workspace packages) from `0.100.13` to `0.100.18`, aligning with the current published engine.
- Updates the template's `bun.lock` accordingly.

## Test plan
- [x] `LINK_LOCAL=1 bun run templates/run-template-tests.ts batcher-validations` → **PASS** (14s, 7 tests).

PR #821, opened 2026-07-31
chore(templates/projected-nft-preorder): bump @effectstream/* to 0.102.0

acedward:
Bumps every `@effectstream/*` dependency in `templates/projected-nft-preorder` from `0.101.1` to the published `0.102.0`, and regenerates `bun.lock`.


acedward:
CI failure here is **pre-existing on v-next**, not caused by this bump.

I pushed a throwaway branch off v-next that touches only this template (keeping 0.101.1) to force a baseline CI run: [run 30643718076](https://github.com/effectstream/effectstream/actions/runs/30643718076) — `projected-nft-preorder` fails there too, with the same error.

The failure is `browserType.launch: Executable doesn't exist at /root/.cache/ms-playwright/chromium_headless_shell-1217/...`. The CI image only runs `playwright install --with-deps chromium` for the version pinned in `e2e/wallets-ui` (.github/Dockerfile:129); templates that run their own `bunx playwright test` never get a matching browser download. Playwright resolves to 1.59.1 on both v-next and this branch, and the lockfile diff contains no playwright or chromium lines.

12 of the 13 template checks pass; only the Playwright-driven ones fail. Worth fixing separately in the CI image.

PR #826, opened 2026-07-31
chore(templates/evm-midnight-v2): bump @effectstream/* to 0.102.0

acedward:
Bumps every `@effectstream/*` dependency in `templates/evm-midnight-v2` from `0.101.1` to the published `0.102.0`, and regenerates `bun.lock`.


acedward:
CI failure here is **pre-existing on v-next**, not caused by this bump.

Baseline run at 0.101.1 on a probe branch off v-next: [run 30643718076](https://github.com/effectstream/effectstream/actions/runs/30643718076) — `evm-midnight-v2` fails there with an identical chain:

```
[compile-evm-contracts-forge] error: modifier-style base constructor call without arguments
[BUILD ERROR] ... rollup/dist/es/shared/node-entry.js:5682:38
[FAIL] Frontend vite build exits successfully
error: Frontend server did not start within 60s
```

The forge compile failure leaves `@evm-midnight/contracts-evm` without a generated entry, so the vite build cannot resolve it and the frontend server never starts. Independent of the version bump; worth fixing separately.

PR #833, opened 2026-07-31
chore(templates/zk-cardano): bump @effectstream/* to 0.102.0

acedward:
Bumps every `@effectstream/*` dependency in `templates/zk-cardano` from `0.101.1` to the published `0.102.0`, and regenerates `bun.lock`.


acedward:
The bump alone broke this template's frontend build (reproduced twice in CI; a 0.101.1 baseline run of the same template passes). Root cause and fix pushed as a second commit.

At 0.101.1, `@effectstream/{batcher-sdk,midnight-contracts,sync}` transitively pulled in the legacy `@midnight-ntwrk/*` wallet-sdk tree, whose `wallet-sdk-unshielded-wallet` re-exported `InMemoryTransactionHistoryStorage`. 0.102.0 drops that legacy tree, so the import in `packages/frontend/client/src/midnight-api.ts` resolves against `@midnightntwrk/wallet-sdk-unshielded-wallet@3.1.0`, which does not export the symbol — rollup fails with an unresolved-export error.

The class actually lives in `@midnightntwrk/wallet-sdk-abstractions` (already a declared dependency, already imported on the adjacent line for `NetworkId`), so the fix moves one named import. Verified locally: `bun run build` in `packages/frontend` now completes (2697 modules, 12.2s).

PR #836, opened 2026-08-01
feat(solana): add the SOLANA:TokenAccount primitive (closes #815 gap 6)

mzf11125:
Closes known gap 6 from #815. Complements #820, which closes gap 1. No file overlap with it.

#815's description named a third Solana primitive, `SolanaTokenAccount`, that was never written. @acedward went looking for it and came up empty, and I confirmed the same across every ref I could reach: `feat/solana-support`, `feat/solana-sync-primitives-batcher`, `solana-clean` and `feat/solana-starter-ci-enable` all carry exactly `solana-account-balance` and `solana-program-log`. So this is written fresh rather than ported.

## What it does

Reads `meta.postTokenBalances`. Narrowed by any combination of `mint`, `owner` and `tokenAccount`, with at least one required. `tokenProgramId` optionally pins it to classic SPL Token or to Token-2022, and omitting it accepts both.

Payload: `{ tokenAccount, mint, owner, amount, decimals, slot }`.

Two properties carried across from #815's security review, both with tests confirmed to fail when the guard is removed:

| | |
|---|---|
| **B3** | `accountIndex` indexes the **resolved** account list, not `message.accountKeys`, so a token account reached through an address lookup table is only findable after `resolveAccountKeys`. |
| **B2** | Reverted transactions are skipped by the existing per transaction guard, so a rolled back mint drives nothing. |

An entry with no filter at all is rejected at construction. Without one it would match every token balance on chain, which is never the intent, and the constructor is the earliest place to catch it.

## The dispatch change was a prerequisite, and not for the reason I first thought

`readPrimitives` dispatched on field truthiness rather than on primitive type:

```ts
if (prim.address)   { /* AccountBalance */ continue; }
if (prim.programId) { /* ProgramLog */ }
```

I originally expected a token primitive to be misrouted into the AccountBalance branch. Reading the actual code, the plain failure is quieter than that: a `TokenAccount` entry matches **no** branch and silently emits nothing. Misrouting only happens if it carries `address`, which `SolanaPrimitive` permits because it is a flat bag of optionals. Neither case is a type error, and both are fixed by dispatching on `prim.type`. That is the first half of #815's known gap 7.

Worth flagging two things about it:

1. **This introduces a pattern rather than following one.** The NEAR fetcher dispatches the same field truthy way (`near/fetcher.ts:178` and `:220`). Three Solana primitives make the field approach ambiguous, so Solana moves first. NEAR is worth converting separately.
2. **`@effectstream/sync` does not depend on `@effectstream/sm`**, so the fetcher could not import `PrimitiveTypeSolana*` from where they lived. The three literals moved into `@effectstream/config`, beside the `SolanaPrimitive` type they discriminate, and `sm/primitives/src/builtin.ts` re-exports them under their existing `PrimitiveType*` names. Every current import site is unchanged, and there is no new dependency and no cycle.

An unrecognised type now warns once per primitive rather than being silently dropped, and once rather than per transaction so catch up does not bury the log.

## Two deliberate choices

**`amount` is a string.** SPL amounts are u64. They do not survive a JS number, and at the top of the range they also exceed PostgreSQL's signed `BIGINT`, so the grammar carries the raw base units as a string and the e2e table uses `TEXT`. This matches `nep141Grammar` and `erc20Grammar`.

**`getConfig()` emits both `stateMachinePrefix` and `scheduledPrefix`.** Every other builtin emits only the latter, which the `Primitive` constructor does not read, so a config round tripped through `getConfig()` silently loses its state machine routing. Emitting both makes this primitive survive that trip and follows the deprecation direction @acedward set in #815.

## Documented limitation

Post state balances only, matching AccountBalance. A token account that is **closed** appears in `preTokenBalances` and is absent from `postTokenBalances`, so closure emits nothing rather than a zero balance. A pre/post delta would cover it and is a deliberate non goal here. It is called out in the primitive docblock and in the chain docs rather than left implicit.

## Verified

| | |
|---|---|
| `bun test packages/node-sdk/sync/src/sync-protocols/solana/` | **42 pass**, up from 31 |
| `bun test packages/node-sdk/sm/primitives/src/solana-token-account/` | **9 pass** |
| `bun test ./packages` | **261 pass**, up from 241. Failure set byte identical to clean `v-next` |
| `bun run e2e/runner.ts solana` | **31 passed, 0 failed**, up from 24 |
| `docs/site` build | clean, only the pre existing blog `#whats-next` anchors |

**No regressions.** I compared the full unit suite against clean `v-next` by stashing, and the 13 failures plus 6 errors are identical in both. They are all pre existing and unrelated: a missing `@midnightntwrk/wallet-sdk-hd`, and runtime reproduction tests that need Postgres and free ports. The pass count moved by exactly the 20 tests added here.

**The e2e assertions are not vacuous.** I re ran the suite capturing the log specifically to check, because a green summary can hide a sync node that never did the work. The state machine fired with real data:

```
[STM] solana-token-account: slot=7 account=8taZEqZ... mint=2KW2XRd9... owner=J2xccRtu... amount=0
[STM] solana-token-account: slot=8 account=8taZEqZ... mint=2KW2XRd9... owner=J2xccRtu... amount=2500000
```

Slot 7 is the ATA opening and slot 8 is the mint, both landing in `solana_token_events` through the full path: `postTokenBalances` to `resolveAccountKeys` to primitive to grammar to STM to Postgres.

**Revert and confirm it fails**, on each new guard, since a guard whose test still passes without it is not a guard:

* drop the `resolveAccountKeys` call, and the lookup table test fails
* simulate the old field truthy dispatch, and both the misroute test and the unsupported type test fail

## Notes

* `@solana/spl-token` is new in `e2e/solana`. The test validator carries SPL Token and the ATA program in genesis, which I verified against the pinned Agave 3.0.14 specifically rather than assuming, since #815 pinned away from 1.18.26.
* The e2e mint and owner come from fixed seeds, using the same idiom as the existing `WATCHED_BALANCE_ADDRESS`, because a static config cannot name a runtime generated address. The first e2e assertion re derives both and checks they match `config.ts`, so a changed seed fails loudly instead of leaving the primitive watching a mint nothing ever touches. Localnet only, and the seeds are public, so they should never be reused anywhere real.
* Gap 3 is deliberately untouched. #820 owns `templates/solana-starter`, so wiring `RateLimitStore` waits for it to merge.
* One local environment note for anyone reproducing this. My `packages/binaries/solana-node/vendor/` still held solana-labs 1.18.26 from before the Agave pin, and the checksum guard from `60c873b5` rejects it, correctly. It fails closed rather than re downloading, so a stale vendor directory has to be cleared by hand.

### DIFF
--- a/.claude/skills/create-effectstream-app/references/chains/solana.md
+++ b/.claude/skills/create-effectstream-app/references/chains/solana.md
@@ -36,6 +36,7 @@ Sync protocol: `SOLANA_RPC_PARALLEL`. Polls slot by slot; **skipped slots are no
 |---|---|---|
 | `PrimitiveTypeSolanaProgramLog` | `programId` (+ optional `eventType` substring) | `{ programId, slot, logMessages }` |
 | `PrimitiveTypeSolanaAccountBalance` | `address` | `{ address, lamports, slot }` |
+| `PrimitiveTypeSolanaTokenAccount` | at least one of `mint` / `owner` / `tokenAccount` (+ optional `tokenProgramId`) | `{ tokenAccount, mint, owner, amount, decimals, slot }` |
 
 Program-log attribution keys off the log stream's `invoke`/`success` framing, not `accountKeys` — a program merely referenced as an account is not treated as invoked, and only its own log lines are forwarded. Reverted transactions are skipped.
 
--- a/bun.lock
+++ b/bun.lock
@@ -445,6 +445,7 @@
         "@effectstream/sm": "workspace:*",
         "@effectstream/utils": "workspace:*",
         "@sinclair/typebox": "^0.34.30",
+        "@solana/spl-token": "^0.4.9",
         "@solana/web3.js": "^1.95.0",
         "bs58": "^6.0.0",
         "effection": "^3.5.0",
@@ -509,7 +510,7 @@
     },
     "packages/batcher": {
       "name": "@effectstream/batcher-sdk",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "dependencies": {
         "@effectstream/crypto": "workspace:*",
         "@effectstream/event-client": "workspace:*",
@@ -554,7 +555,7 @@
     },
     "packages/binaries/avail-light-client": {
       "name": "@effectstream/npm-avail-light-client",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "npm-avail-light-client": "index.js",
       },
@@ -565,7 +566,7 @@
     },
     "packages/binaries/avail-node": {
       "name": "@effectstream/npm-avail-node",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "npm-avail-node": "index.js",
       },
@@ -576,7 +577,7 @@
     },
     "packages/binaries/bitcoin-core": {
       "name": "@effectstream/bitcoin-core",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "bitcoin-core": "./index.js",
       },
@@ -586,7 +587,7 @@
     },
     "packages/binaries/celestia": {
       "name": "@effectstream/celestia",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "celestia": "./index.js",
       },
@@ -596,7 +597,7 @@
     },
     "packages/binaries/grafana-alloy": {
       "name": "@effectstream/grafana-alloy",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "grafana-alloy": "index.js",
       },
@@ -606,7 +607,7 @@
     },
     "packages/binaries/grafana-loki": {
       "name": "@effectstream/grafana-loki",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "grafana-loki": "index.js",
       },
@@ -616,7 +617,7 @@
     },
     "packages/binaries/midnight-indexer": {
       "name": "@effectstream/npm-midnight-indexer",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "npm-midnight-indexer": "index.js",
       },
@@ -629,7 +630,7 @@
     },
     "packages/binaries/midnight-node": {
       "name": "@effectstream/npm-midnight-node",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "npm-midnight-node": "index.js",
       },
@@ -641,7 +642,7 @@
     },
     "packages/binaries/midnight-proof-server": {
       "name": "@effectstream/npm-midnight-proof-server",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "npm-midnight-proof-server": "index.js",
       },
@@ -652,7 +653,7 @@
     },
     "packages/binaries/near-sandbox": {
       "name": "@effectstream/near-sandbox",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "near-sandbox": "./index.js",
       },
@@ -662,7 +663,7 @@
     },
     "packages/binaries/ord": {
       "name": "@effectstream/ord",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "ord": "./index.js",
       },
@@ -672,7 +673,7 @@
     },
     "packages/binaries/solana-node": {
       "name": "@effectstream/solana-node",
-      "version": "0.100.18",
+      "version": "0.102.0",
       "bin": {
         "solana-node": "./index.js",
       },
@@ -707,7 +708,7 @@
     },
     "packages/build-tools/orchestrator": {
       "name": "@effectstream/orchestrator",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "bin": {
         "orchestrator": "src/cli.ts",
       },
@@ -717,11 +718,11 @@
     },
     "packages/chains/avail-contracts": {
       "name": "@effectstream/avail-contracts",
-      "version": "0.101.1",
+      "version": "0.102.0",
     },
     "packages/chains/bitcoin-contracts": {
       "name": "@effectstream/bitcoin-contracts",
-      "version": "0.101.1",
+      "version": "0.102.0",
       "dependencies": {
         "bitcoinjs-lib": "^7.0.0",
         "ecpair": "^3.0.0",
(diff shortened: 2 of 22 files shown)


mzf11125:
@acedward when you have a moment, two asks on this one.

**Review**, whenever it suits. This closes known gap 6 from #815, the SPL token primitive that was described in #759 and never written. Your pointers were all exactly where you said, so it was additive rather than archaeology.

**CI, which this PR cannot get on its own.** `.github/workflows/main.yaml` is `on: [push]` with no `pull_request` trigger, so a push to a fork fires the workflow inside the fork and nothing upstream is triggered by opening the PR. Same thing that happened to #759. If you want it validated in repo, re pushing the branch the way #759 became #815 is the reliable route, and I am happy for you to take it over that way.

Recording the runs so the work is not invisible in the meantime.

## Rebased onto current `v-next`

Originally cut from `47ffd0c7`, now rebased onto `96518069` and force pushed, so it carries the 38 commits that landed since, including the node20 Actions deprecation fix in #834. That last one is why I re ran CI rather than pointing at the earlier run. Clean rebase, no conflicts, not even in `bun.lock`. PR reports `MERGEABLE / CLEAN`.

## CI

[Run 30699531695](https://github.com/mzf11125/effectstream/actions/runs/30699531695) @ `cc0f2e80`, the rebased head, **fully green**:

```
All 11 suite(s) passed        [PASS] solana (104.2s)
```

| job | result |
|---|---|
| `changes` | success |
| `template-tests` | skipped, correctly, this branch touches no template |
| `e2e` | success, 29m 22s |
| `ci-ok` | success |

Every one of the 11 suites passed: sync-repro, evm, bitcoin, cardano, midnight, avail, celestia, near, solana, features, wallets. Solana reports **31 tests passed, 0 failed**, up from 24. That breadth is the part that matters here, because the `prim.type` dispatch change touches the shared Solana fetcher path and not only the new primitive.

[Run 30698429429](https://github.com/mzf11125/effectstream/actions/runs/30698429429) @ `6dd3cad9` was the same tree on the pre rebase base, also green across all 11, so both bases are covered.

Worth saying plainly: **the workflow references no secrets**, so a fork run is the same jobs on the same code rather than a degraded proxy. The only thing missing is the reporting link back to this PR.

## The token primitive did real work, not a vacuous pass

Checked on purpose, since a green summary can hide a sync node that never executed the path. From the CI e2e log:

```
[STM] solana-token-account: slot=6 account=8taZEqZ... mint=2KW2XRd9... owner=J2xccRtu... amount=0
[STM] solana-token-account: slot=7 account=8taZEqZ... mint=2KW2XRd9... owner=J2xccRtu... amount=2500000
```

Slot 6 is the associated token account opening and slot 7 is the mint, both travelling `postTokenBalances` to `resolveAccountKeys` to primitive to grammar to state machine to Postgres. The solana suite reports **31 passed, 0 failed**, up from 24.

## Two things worth your eyes in review

1. **The dispatch change was a prerequisite, and my first framing of it was wrong.** I expected a token primitive to be misrouted into the AccountBalance branch. Reading the code, the plain failure is quieter: a TokenAccount entry matches no field branch and silently emits nothing. Misrouting only happens if it carries `address`, which the flat optionals permit. Both are fixed by dispatching on `prim.type`, which is the first half of your gap 7. It changes behaviour for all three Solana primitives, not only the new one, so it is the part most worth a second pair of eyes.

2. **`getConfig()` emits both `stateMachinePrefix` and `scheduledPrefix` here**, where every other builtin emits only the latter. It follows the deprecation direction you set, and it stops a config round tripped through `getConfig()` from silently losing its state machine routing, but it is a one line divergence from twenty siblings. If you would rather that were handled repo wide instead of per primitive, say so and I will pull it out.

Also noting #820 merged, so gap 3 is unblocked. I will pick up `RateLimitStore` in the template next unless you are already on it.

acedward:
Thanks for the contribution :)
Review: All Good!


On the two items you flagged:

1. The dispatch change: agree, including moving the literals into @effectstream/config. The re-export keeps every import site stable, and I confirmed there's no new edge in the dependency graph. NEAR converting separately is right; this PR shouldn't grow. 
2. getConfig() emitting both prefixes, lets keep it. It's the documented deprecation direction and it fixes a real round-trip loss. converting the other twenty builtins is a separate mechanical PR.

acedward:
I will fix the cicd - and will release this soon, I got other PRs to merge. 
There is a [hackathon](https://hackbuenosaires.com/) this week and we will be showing this new solana integration to the participants as well :)

acedward:
0.103.0 published with this change :)

mzf11125:
Thanks, and good to see it out already.

I pulled 0.103.0 from npm rather than assuming, and the change is intact across all three packages:

- `@effectstream/config` carries `SOLANA_PRIMITIVE_TOKEN_ACCOUNT` at `src/schema/sync-protocols/types.ts:170`, so the shared discriminator moved as intended
- `@effectstream/sync` has `switch (prim.type)` at `fetcher.ts:323` with the token case at `:389`
- `@effectstream/sm` ships the whole `primitives/src/solana-token-account/` directory

Docs landed too, `211-solana.md` has the Token Account section and `118-primitives.md` has the table row. Nothing to do on either of the two items you ruled on, both stay as they are.

## One thing worth checking before the hackathon

Since you are putting this in front of participants this week, this is worth five minutes.

Rate limiting is already on by default for every batcher, `templates/solana-starter` included. `batcher-server.ts:189` falls back to `DEFAULT_CONFIG_VALUES.rateLimit` and `:190` falls back to `InMemoryRateLimitStore` when the config omits them, and the template's `batcher.dev.ts` sets no `rateLimit`. So the template is not unprotected, which is the good news.

The defaults are **1000 requests per rolling 24 hour window**, and no adapter in the repo implements `getRateLimitKeyStrategy()`, so `batcher-server.ts:323` resolves to `"ip"` for all of them. One bucket per IP address.

At a venue, participants are usually behind one NAT. If you host a shared batcher for the event, the whole room draws down a single `ip:` bucket of 1000, on a window that does not reset at any point during the day. It just starts answering `429` with a `Retry-After`. That is a quiet way for a live demo to stop working, and to the room it looks like the batcher is broken rather than like a limit being reached.

Two small mitigations if you want them before I get anything merged:

```ts
rateLimit: { maxRequests: 100000, windowMs: 60_000 }
```

on the event batcher config, and/or a `getRateLimitKeyStrategy()` returning `"ip-and-address"` on the Solana adapter, so the per wallet key does the real work instead of the shared IP.

Related small bug while I was in there. `RateLimitConfig` at `packages/batcher/core/config.ts:124` and `:126` documents the defaults as 100 requests and 60000 ms, but `DEFAULT_CONFIG_VALUES.rateLimit` at `:192` is 1000 and 86400000. The jsdoc is the wrong one, and `templates/batcher-validations/README.md` already states the real values. Anyone reading the interface to decide whether they need to override it sees a window 1440 times shorter than reality.

## What I am picking up next

Working down the rest of the #815 list, smallest and most useful first, as separate PRs rather than one:

- **Gap 3, the rate limit wiring.** Explicit config in `solana-starter`, the adapter key strategy, and the jsdoc correction, for the reasons above.
- **Gap 5, checksums for the other binaries.** `solana-node` is currently the only one of twelve `packages/binaries/*` that verifies what it downloads. The other eleven take whatever the URL returns.
- **The rest of gap 7.** The dispatch half went in with #836. Still open are `Output.transactions` being buffered per block with no consumer, and `readData` doing one serial RPC round trip per slot.

Gaps 2 and 4 read like issues rather than code to me, so I will file them as issues unless you would rather they stay in the #815 description.

Shout if any of these is already yours or you want a different order, especially while the event is running.

PR #845, opened 2026-08-03
fix(batcher): make the rate limit an explicit, per-wallet spend cap (closes #815 gap 3)

mzf11125:
Closes known gap 3 from #815 for the SDK and docs. Follow up to #836.

## The gap was described slightly wrong, and the correction matters

Gap 3 reads "no rate limit". There is one. `batcher-server.ts:189` falls back to `DEFAULT_CONFIG_VALUES.rateLimit` and `:190` to `InMemoryRateLimitStore` whenever a config omits the block, so **every batcher is rate limited already**, including `templates/solana-starter`. Nothing was unprotected.

The actual problems are two, and both are quiet:

1. **The effective values are invisible.** A template with no `rateLimit` key looks unlimited and is in fact 1000 requests per rolling 24 hours. Nobody chose that number.
2. **Keying is per IP only.** `getRateLimitKeyStrategy()` has been declared optional on `adapters/adapter.ts:164` since it was written, and **no adapter in the repo implements it**, so `:323` resolves to `"ip"` for all of them. Everyone behind one NAT shares one bucket. On a shared network that is the normal case, not the edge case, and the failure mode is honest users throttling each other into `429` while no individual is over budget.

Since this is a sponsored batcher, the limit is a **spend cap**, not just abuse control. Every accepted transaction costs the sponsor the 5000 lamport base fee.

## Changes

**`SolanaAdapter` gains `rateLimitKeyStrategy`** and implements the hook. Still defaults to `"ip"`, so no existing deployment changes behaviour. `"ip-and-address"` adds an independent per wallet budget.

That is only sound because `verifySignature` rejects an `input.address` that did not actually sign, so the address key cannot be forged to mint fresh buckets. Worth stating explicitly, because the same option on an adapter without that binding would be a way around the limit rather than a refinement of it. The docblock says so.

**Two conflicting default sets, now one.** The `RateLimitConfig` jsdoc claimed 100 requests per 60000 ms, and the TypeBox schema encoded the same, while `DEFAULT_CONFIG_VALUES` (the one the running server reads) is 1000 per 86400000. A window **1440 times longer** than documented.

This was not only a comment problem. The schema defaults are reachable by a different path, `Value.Cast` filling a partial `rateLimit` object, so the effective limit depended on whether the key was absent or half filled. Both now match the constant, with a comment saying they must.

**`buildRateLimitKeys` is exported** so the strategies can be tested. The choice of keys is the entire behavioural difference between them and is otherwise unobservable without standing up a server and exhausting a limit.

**The template sets `rateLimit` explicitly** with env overridable values, so the number is chosen rather than inherited, and logs the effective limit at boot because a `429` in the wild is otherwise hard to attribute.

## What this PR deliberately does not do

`templates/solana-starter` **cannot adopt `rateLimitKeyStrategy` yet.** It pins `@effectstream/batcher-sdk` 0.102.0 and `template-tests` installs from npm, so referencing an option that only exists in this branch would break that job. I checked the published tarball rather than assuming: 0.102.0 exports `InMemoryRateLimitStore` (so the explicit block works today) and has no `rateLimitKeyStrategy` (so the strategy cannot). The template carries a comment pointing at the follow up instead.

Same shape as #820 waiting for 0.102.0. One line in `templates/solana-starter/packages/batcher/solana-adapter.ts` after the next release finishes gap 3.

## Verified

| | |
|---|---|
| `bun test packages/batcher` | **56 pass, 0 fail** (8 new) |
| `bun test ./packages` | **302 pass** vs **294** on a stashed baseline. Exactly +8, identical 4 fail / 2 errors both ways, so no regression from the exported function or the schema change |
| `templates/run-template-tests.ts solana-starter` | **14 passed, 0 failed**, the real CI path, against published 0.102.0 |
| docs build | passes. The `#whats-next` blog anchor warnings are pre-existing |

**Live `429`, because a rate limit that is only unit tested is not a rate limit.** Booted the template batcher with `maxRequests=3` and posted five schema valid requests:

```
ratelimit: 3 req / 60000 ms, keyed by ip
req 1 -> HTTP 401   Invalid signature
req 2 -> HTTP 401   Invalid signature
req 3 -> HTTP 401   Invalid signature
req 4 -> HTTP 429   retry-after: 60   Rate limit exceeded
req 5 -> HTTP 429   retry-after: 60   Rate limit exceeded
```

Two things fall out of that, beyond the limit working. The explicit config is genuinely read, not silently ignored. And **the limiter runs before signature verification**, so the three rejected requests still consumed budget. That is the right posture for a spend gate, since an attacker should not get unlimited free attempts by sending garbage, but it is worth knowing when sizing `maxRequests`.

I also reverted the strategy plumbing and confirmed the new adapter test fails. A guard whose test still passes without it is not a guard.

## Note on CI

Same situation as #836. `.github/workflows/main.yaml` is `on: [push]` with no `pull_request` trigger, so opening this from a fork triggers nothing upstream. The workflow references no secrets, so a fork run is the same jobs on the same code. I will post the fork run once it finishes. Re pushing in repo is the reliable route if you want it reported here.

@acedward flagging the venue case specifically because of the hackathon. If you are running a shared batcher this week, the two things worth doing to it are an explicit `rateLimit` sized to your funding and `rateLimitKeyStrategy: "ip-and-address"`, and neither needs this PR merged.


mzf11125:
Recording the full local verification, since this branch cannot report CI back to this PR on its own.

Branch is on top of current `v-next` (`f1247f32`, which includes #844), zero commits behind, so none of this is against a stale base.

## Local test runs

**Batcher unit tests**

```
$ bun test packages/batcher
 56 pass
 0 fail
 87 expect() calls
Ran 56 tests across 6 files. [610.00ms]
```

8 of those are new: 5 on `buildRateLimitKeys`, 2 on the adapter strategy, 1 asserting that omitting `rateLimit` does not mean unlimited.

**Full unit suite, against a stashed baseline of the same tree**

|  | pass | skip | fail | errors |
|---|---|---|---|---|
| `v-next` baseline (`git stash push -u`) | 294 | 2 | 4 | 2 |
| this branch | **302** | 2 | 4 | 2 |

Exactly `+8`, which is the new tests and nothing else. Identical failure set both ways, so exporting `buildRateLimitKeys` and changing the schema defaults regressed nothing. The 4 failures and 2 errors are pre-existing on `v-next` and unrelated.

**Template suite, the real CI path**

```
$ bun run templates/run-template-tests.ts solana-starter
Results: 14 passed, 0 failed
  [PASS] solana-starter (3m42s)
  All 1 template(s) passed
```

This is the run that matters most for the pinning question. It installs `@effectstream/batcher-sdk` **0.102.0 from npm**, so it proves the explicit `rateLimit` block works against the published SDK rather than only against the workspace.

**Docs build**

```
$ cd docs/site && bun run build
[SUCCESS] Generated static files in "build".
```

The `#whats-next` blog anchor warnings are pre-existing on `v-next` and unrelated to this change.

## Live 429, end to end

A rate limit that is only unit tested is not a rate limit. Booted the template batcher with `maxRequests=3` and posted five schema valid requests to `/send-input`:

```
  ratelimit: 3 req / 60000 ms, keyed by ip

req 1 -> HTTP 401   Invalid signature
req 2 -> HTTP 401   Invalid signature
req 3 -> HTTP 401   Invalid signature
req 4 -> HTTP 429   retry-after: 60   Rate limit exceeded
req 5 -> HTTP 429   retry-after: 60   Rate limit exceeded
```

Three things this establishes that the unit tests cannot:

1. The explicit config is actually read. The startup line reflects the env overrides, and the cutoff lands at request 4 for `maxRequests=3`.
2. `Retry-After` is set on the response, not only the JSON body.
3. **The limiter runs before signature verification.** The three rejected requests still consumed budget. That is the correct posture for a spend gate, since garbage submissions should not be free and unlimited, but it is worth knowing when sizing `maxRequests`: your budget is consumed by rejected traffic too.

## Negative control

Reverted the strategy plumbing (`config.rateLimitKeyStrategy ?? "ip"` back to a hardcoded `"ip"`) and confirmed the new test fails rather than passing anyway:

```
error: expect(received).toBe(expected)
Expected: "ip-and-address"
Received: "ip"
(fail) SolanaAdapter honours a configured rate limit strategy

 19 pass
 1 fail
```

Then restored it and reran clean. A guard whose test still passes without the guard is not a guard.

## Fork CI: fully green

[Run 30794284517](https://github.com/mzf11125/effectstream/actions/runs/30794284517) @ `4527d1c8`, every job success.

| job | result |
|---|---|
| `changes` | success |
| `assets-check` | success |
| `template-tests` | success, 24m 54s |
| `e2e` | success, 35m 49s |
| `ci-ok` | success |

**`template-tests` ran rather than skipping**, because this branch touches `templates/solana-starter/**`. That is the job that would have caught the pinning problem, so it is the one that mattered here. All 14 templates passed, `solana-starter` among them:

```
[PASS] solana-starter (1m34s)     [PASS] batcher-validations (22s)
```

`batcher-validations` passing is worth calling out separately. Its README documents the rate limit defaults, and it exercises them rather than configuring them, so it is the closest thing the repo has to a regression test on the numbers I changed.

**e2e, all 11 suites passed:**

```
evm 34.0s      bitcoin 130.1s   cardano 106.5s   midnight 1119.0s
avail 202.0s   celestia 38.4s   near 12.6s       solana 100.5s
features 59.8s wallets 56.6s    sync-repro 91.9s
All 11 suite(s) passed
```

That breadth is what I wanted from CI here. The schema default change and the exported `buildRateLimitKeys` sit in shared batcher code that every chain's batcher path runs through, so a green solana suite alone would not have been enough.

mzf11125:
@acedward this one is ready for you whenever it suits.

**CI is fully green**, all six jobs: [run 30794284517](https://github.com/mzf11125/effectstream/actions/runs/30794284517) @ `4527d1c8`. `e2e` all 11 suites, `template-tests` all 14 templates. Details and the local runs are in the comment above.

Two things worth your attention rather than just an approve:

1. **The gap 3 description was wrong and I changed the fix accordingly.** It reads "no rate limit", but there is one: the server falls back to 1000 requests per 24 hours and an in-memory store whenever a config omits the block, so nothing was ever unprotected. What was actually broken is that the number was invisible and keying was per IP only. If you disagree with that reading, the whole shape of this PR is the thing to push back on.

2. **The template only gets half of it, on purpose.** `solana-starter` pins batcher-sdk 0.102.0 and `template-tests` installs from npm, so it cannot reference `rateLimitKeyStrategy` until that ships. I verified against the published 0.102.0 tarball rather than assuming: `InMemoryRateLimitStore` is exported, the new option is not. One line in `templates/solana-starter/packages/batcher/solana-adapter.ts` finishes gap 3 after the next release.

**Independent of this PR, for the hackathon.** If you are hosting a shared batcher this week, the venue case is worth five minutes regardless of whether this merges: a room behind one NAT shares a single 1000 request bucket on a window that does not reset during the day, and it fails as `429` that reads like a broken batcher. An explicit `rateLimit` sized to your funding, plus `"ip-and-address"` keying, both work on 0.103.0 today.

Gap 5 is up as #846 as well, if you would rather review them together.

acedward:
thanks for the PR, im a little behind on this

acedward:
Updated this PR to address six confirmed review issues:

- Added an atomic `globalMaxRequests` ceiling scoped per adapter target.
- Moved identity/global quota consumption after signature verification, preventing forged-address bucket poisoning.
- Fixed shared-NAT behavior: verified wallets receive independent identity allowances until the target-global ceiling is reached.
- Replaced the race-prone count-then-hit store contract with atomic multi-bucket `RateLimitStore.consume`.
- Corrected Solana fee guidance: 5,000 lamports per signature, meaning sponsored transactions cost at least 10,000 lamports; 100k transactions therefore cost at least 1 SOL.
- Updated template startup diagnostics to report normalized runtime values and documented the pinned SDK versus `LINK_LOCAL=1` behavior.

Verification:

- Batcher tests: 81 passed, 0 failed.
- Linked Solana starter integration: 14 passed, 0 failed.
- HTTP regression confirms forged request `401`, genuine request `200`, and exhausted quota `429` with `Retry-After`.
- The remaining full-suite failures reproduce unchanged on the current `v-next` baseline.

Note: custom `RateLimitStore` implementations must migrate to the new atomic `consume(buckets, nowMs, windowMs)` contract.

acedward:
Merge-readiness update for final head `b66a5b16e4f727d4c0b8d7e518faad754294759e`.

`b66a5b16` has the same tree as security-fix commit `2c8d01882282600d288d50faa8c31c73c35398c2`; the empty follow-up commit only retriggered fork CI because this fork token cannot rerun failed jobs.

Key issues fixed:

- Integrated current `v-next` (`f143af9f`), including the Solana template `0.103.3` dependency/base changes and corrected template-runner behavior.
- Restored a pre-authentication abuse budget keyed only by trusted transport IP and consumed before adapter lookup/signature verification. Attacker-controlled `target`, `address`, signature, and payload fields cannot mint new pre-auth buckets.
- Kept target-global and verified-identity spend budgets after successful signature verification, so forged requests cannot poison a victim wallet or sponsor-global bucket.
- Made each phase's multi-bucket admission atomic and deduplicated, including the concurrent target-global cap.
- Replaced hardcoded SDK-version diagnostics with capability-based reporting for published-SDK and `LINK_LOCAL=1` modes.

Security/diff audit:

- The final diff is confined to 23 expected batcher source/test, documentation, and Solana/batcher template files.
- No dependency manifest or lockfile changes relative to current `v-next`; no workflow changes, executable modes, binaries, symlinks/submodules, download commands, new process spawning, or embedded secrets.
- GitHub currently reports the final SHA as `MERGEABLE` / `CLEAN`, and current `origin/v-next` remains `f143af9f`.

Isolated test evidence:

- Focused layered-limit tests: 38 pass, 0 fail, 84 assertions.
- Full batcher suite: 85 pass, 0 fail, 162 assertions.
- Full packages candidate/baseline comparison: candidate 336 pass / 2 skip / 4 fail; exact-base baseline 316 pass / 2 skip / the same 4 failures. No candidate-only package failure.
- Documentation sync: 39 package pages and 18 template pages current. Candidate and baseline builds reach the same pre-existing Docusaurus MDX `empty preset` failure.
- Published SDK Solana template: 14/14 pass.
- `LINK_LOCAL=1 SKIP_SOLANA_BUILD=1` Solana template: 14/14 pass.
- Untrusted code was executed only in disposable Docker isolation. The only published endpoints were bound to loopback on random host ports `55001` and `55002` (>10000); test containers/images/volumes/networks were removed afterward.

Exact-head fork CI: [run 31525637664](https://github.com/mzf11125/effectstream/actions/runs/31525637664)

- PASS: `changes`, `assets-check`, `frontend-build`, and `e2e` (36m25s).
- The changed `solana-starter` template passed 14/14 in CI.
- RED: aggregate `template-tests` because unrelated `cardano-delegation` setup passed `-` to `cardano-cli --tx-in`, and `preorder` timed out waiting for Vite after all 45 backend tests passed. These same two failures reproduced on the prior exact-content run. The official workflow for the exact base is also red in unrelated template tests, including the identical `preorder` timeout: [base run 31513883141](https://github.com/effectstream/effectstream/actions/runs/31513883141).
- Consequently `ci-ok` is red. I could not rerun only failed jobs because GitHub requires fork admin rights.

**BREAKING:** custom `RateLimitStore` implementations must migrate from split `count`/`hit` operations to atomic `consume(buckets, nowMs, windowMs)`. The old interface is intentionally incompatible because it cannot enforce multi-bucket admission without races. Redis implementations should use one transaction/Lua script; SQL implementations should use transactional locking. A split compatibility fallback must not be added.

**Current recommendation: do not merge yet.** Although the changed code and relevant tests are clean, the plan requires both (1) a trusted maintainer/code owner to independently review and explicitly approve this exact final SHA, because the original PR source is untrusted, and (2) an admin rerun that gets required `template-tests` and aggregate `ci-ok` green. Please review `b66a5b16` and rerun the failed job when possible.

acedward:
Final update after merging current `v-next`:

- Final PR head: `aa357c7a36808cf20d88708cec543d7cc7560185`
- Current base: `093006ed98f969aee862c0d95d07b4d37b2784e4`
- The branch now includes the external-PR CI fix and the follow-up template/e2e infrastructure fixes from `v-next`.
- The PR-relative binary patch stayed byte-for-byte identical across both base catch-ups: `b62868b8aa28575498437d9c0f9e8a38dde6b0e0b2d2649b96aa94837f4469b4`.
- Final diff remains the same 23 expected `100644` batcher/docs/Solana-template paths, with no PR-added workflows, dependency manifests/locks, executable modes, binaries, symlinks/submodules, secrets, downloads, or process-spawn additions.

Authoritative upstream CI is fully green on the exact final head: [run 31536206055](https://github.com/effectstream/effectstream/actions/runs/31536206055). `changes`, `assets-check`, `frontend-build`, `template-tests`, `e2e`, and aggregate `ci-ok` all passed. GitHub reports the PR `MERGEABLE` / `CLEAN` and current with `v-next`.

Key issues fixed:

1. Unauthenticated abuse is limited by a trusted transport-IP budget before signature verification; attacker-controlled target/address/signature/payload cannot mint new pre-auth buckets.
2. After verification, target-global and verified-identity budgets are consumed atomically, preventing forged-wallet poisoning and concurrent sponsor-cap overshoot.
3. Shared-NAT fairness and victim eligibility are covered by regression tests; invalid signatures never consume claimed-wallet or sponsor-global budgets.
4. Solana/template diagnostics are capability-based and accurately distinguish published-SDK and `LINK_LOCAL=1` behavior.
5. Documentation and generated copies describe the layered model, in-memory limitations, and upstream WAF requirement.

**Breaking-change warning:** custom `RateLimitStore` implementations must migrate from split `count`/`hit` methods to atomic `consume(buckets, nowMs, windowMs)`. Redis implementations should use a Lua script/transaction; SQL implementations need transactional locking. A split compatibility fallback would reintroduce the concurrency bug.

Because the original contribution is untrusted and there are still zero submitted reviews, my recommendation remains: **do not merge until a trusted maintainer or code owner independently reviews and approves exact head `aa357c7a`**. The code, base-sync, and CI gates are green; the independent trust gate is the only remaining blocker.

PR #846, opened 2026-08-03
chore(binaries): verify downloads in 3 more wrappers, with real provenance (#815 gap 5)

mzf11125:
Advances known gap 5 from #815. Independent of #845, no overlapping files.

`solana-node` was the only one of twelve `packages/binaries/*` that verified what it downloaded before executing it. The other eleven ran whatever the URL returned. This covers three more and puts them all on one implementation.

## The part worth reviewing is what a pinned digest actually proves

Not the hashing. The provenance.

`bin-wrapper` discards the archive after extracting, so a wrapper can only verify the **extracted binary**. Upstream projects publish sums for the **archive**. Those are different artifacts, so you cannot just paste the vendor's number in.

So the generator builds a chain: fetch the archive, match it against the vendor's published `SHA256SUMS`, and only then extract and hash the binary. That makes the recorded digest traceable to something the vendor signed. All **14 new digests** across the three packages are `upstream-verified` this way.

**`solana-node`'s three existing digests are labelled `self-recorded`, not upstream-verified.** They were hashed from a download on one maintainer's machine. That pins the artifact against *later* mutation and attests nothing about whether it was right to begin with. Both are worth having, but they are not the same claim, and relabelling the weaker one as the stronger one would have been the easy wrong move. Provenance is recorded per entry so a reader can tell them apart.

## Changes

**`@effectstream/binary-checksum`**, a new package holding the check. Dependency-free plain ESM, because every wrapper runs under bare `node index.js` with no build step. Keeps the set-membership rationale that `solana-node` documented (matching a set rather than looking up by `os.arch()`, because `arch@2.x` had no notion of arm64 and every Mac silently ran x86_64 under Rosetta).

An empty checksum table now **throws** rather than accepting anything. A forgotten regeneration after a version bump would otherwise quietly turn a fail-closed guard into a no-op, which is the exact failure the module exists to prevent.

**`scripts/generate-binary-checksums.ts`** replaces the hand-run instruction that used to sit in a docblock:

```
shasum -a 256 packages/binaries/solana-node/vendor/bin/solana-test-validator
```

That only ever produced a digest for the machine you happened to be on, which is precisely why the table was never filled in for other platforms. The script downloads every platform's asset, so a full table is now a command rather than a coordination problem. `--check` reports drift.

**Ordering, documented at each call site.** `bin.run()` resolves the binary **and executes it**, so verifying after it protects nothing. `bitcoin-core` is worse: `adhocSignIfNeeded` rewrites the Mach-O on Apple Silicon to satisfy the kernel's exec requirement, which changes the hash. So there verification is a download-time check only and must not be moved below the signing call. There is a comment saying so, because it is the kind of thing a later refactor reorders innocently.

**Incidental fix:** `grafana-loki` and `grafana-alloy` were missing `"type": "module"` while all five sibling packages have it. They use ESM syntax, so node was reparsing them with a warning and a stated performance penalty. Pre-existing, and my new import is what made it audible.

## Verified by running the binaries, not just unit tests

A checksum guard that has only been unit tested has not been shown to sit in the real code path.

| package | result |
|---|---|
| `bitcoin-core` | `verified bitcoind v28.1 (linux-x64)`, regtest started |
| `grafana-loki` | `verified loki v3.5.8 (linux-x64)`, server up on 3100 |
| `grafana-alloy` | `verified alloy v1.11.3 (linux-x64)`, launched |
| `solana-node` | `verified` the already-vendored Agave 3.0.14 as `linux-x64` |

The generated digests match what **bin-wrapper's own download path** extracts. That is the bit that makes this more than the script agreeing with itself: two independent code paths, and separately my throwaway probe script, arrived at the same hashes.

Also:

- Appended a single byte to a copy of the validator and confirmed rejection.
- `bun test ./packages`: **301 pass** vs **294** on `v-next`. Exactly `+7`, the new tests, identical 4 fail / 2 errors both ways.
- **The archive-mismatch guard fired for real.** A truncated cached download (9 KB standing in for 117 MB, left behind by an interrupted transfer) was caught by the upstream comparison rather than being silently hashed and pinned. It was still a structurally valid zip, so "it extracted" proved nothing. That is why the script now checks `content-length` and renames atomically, so a partial file cannot reach the cache under the real name.
- Docs build passes, 39/39 package READMEs sync. The `/docs/home/templates/` broken links are **pre-existing on `v-next`**, unrelated to this: the committed index links `./minimal.md` while the file is `1210-minimal.md`. Probably worth a separate fix, from the recent template page generation.

## What this deliberately does not cover

`ord`, `celestia`, `near-sandbox`, both `avail-*` and all three `midnight-*` still download unverified. Not laziness, and not only download volume:

- **`ord` publishes no `SHA256SUMS`.** Any digest would be self-recorded.
- **`celestia` and `midnight-node` download from `effectstream/binaries`, not upstream.** A digest there pins the re-hosted copy. It does not prove the artifact matches what Celestia or Midnight released, because the re-hosting step is where that would need checking.

Those need a decision about what we are willing to claim, and for the re-hosted ones arguably a change to how artifacts get into `effectstream/binaries` in the first place. Happy to take that on, but it is a different conversation from spreading a mechanical pattern, so I would rather ask than guess.

## Note on CI

Same as #845 and #836. `main.yaml` is `on: [push]` with no `pull_request` trigger, so this cannot report CI to this PR from a fork. I will post the fork run when it finishes.

Worth flagging for review: I did **not** add a checksum-drift job to CI. `--check` has to download every asset on every platform, which is roughly a gigabyte, and that is a poor trade against a table that only changes on a version bump. Regenerating at bump time is the intended workflow. Say the word if you would rather it were gated.


mzf11125:
@acedward this is the gap 5 half, independent of #845 and touching no shared files.

## CI

[Run 30797760635](https://github.com/mzf11125/effectstream/actions/runs/30797760635) @ `224e07f4`.

| job | result |
|---|---|
| `changes` | success |
| `assets-check` | success |
| `template-tests` | skipped, correctly, this branch touches no template |
| `e2e` | in progress |

`changes` passing is the one I was watching, since this adds the **39th publishable package** and I wanted to see the workspace glob pick it up without any registration. It did, and `sync-package-readmes.ts` went from 38/38 to 39/39 on its own. I will update this comment when `e2e` lands.

## Local verification

| | |
|---|---|
| `bun test ./packages` | **301 pass** vs **294** on `v-next`, exactly `+7`, identical 4 fail / 2 errors both ways |
| `bitcoin-core` | `verified bitcoind v28.1 (linux-x64)`, regtest started |
| `grafana-loki` | `verified loki v3.5.8 (linux-x64)`, server up on 3100 |
| `grafana-alloy` | `verified alloy v1.11.3 (linux-x64)`, launched |
| `solana-node` | verified the already-vendored Agave 3.0.14 as `linux-x64` |
| tamper check | appended one byte to a copy of the validator, rejected |
| docs build | passes, 39/39 READMEs synced |

I ran each binary rather than only unit testing, because a guard that has only been unit tested has not been shown to sit in the real code path. The generated digests match what bin-wrapper's own download path extracts, so two independent paths agree on the hashes.

## Three things I would rather you looked at than approved

1. **Provenance is the actual content of this PR.** bin-wrapper discards the archive, so wrappers can only verify the extracted binary, while vendors publish sums for the archive. Different artifacts. The generator chains them: match the archive against the vendor's `SHA256SUMS`, then extract and hash. All 14 new digests are `upstream-verified` that way. `solana-node`'s existing three are labelled `self-recorded` because nobody ever matched them against an Agave-published sum, and calling them upstream-verified would have claimed a guarantee that does not exist.

2. **I stopped at three packages rather than doing all eleven.** `ord` publishes no `SHA256SUMS`. `celestia` and `midnight-node` download from `effectstream/binaries` rather than upstream, so a digest there pins the **re-hosted copy** and proves nothing about what Celestia or Midnight actually released. Making that meaningful probably means checking artifacts as they enter that repo, which is a change to your release process and not mine to assume. Tell me which way you want it and I will do the rest.

3. **No CI drift job.** `--check` has to download every asset for every platform, roughly a gigabyte, against a table that only changes on a version bump. I judged that a bad trade and made regeneration the bump-time workflow instead. Easy to gate if you disagree.

## Unrelated, found on the way

The templates docs index links `./minimal.md` but the file is `1210-minimal.md`, so those links are broken on `v-next` today. Same for the sibling template pages. Looks like it came in with the recent template page generation. Not touched here, happy to send a one line fix separately.

